Issue #2, opened 2023-08-21
set up repo


juliasloan25:
Set up initial state/infrastructure of the repo

Closed by #3 

## To Do
- [x] update README.md #1 
- [x] set up folder structure (.dev, .github, docs, experiments, src, test)
- [x] add project.tomls for all directories except src/
- [x] set up top level project
- [x] copy .dev, .github contents from another CliMA repo
  - [x] include github actions, juliaformatter, compathelper

PR #3, opened 2023-09-08
set up repo

juliasloan25:
## Purpose 
Generic setup for the new repo, including folder structure and Project.toml files.

Closes #2

----

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,4 +21,5 @@ docs/site/
 # It records a fixed state of all packages used by the project. As such, it should not be
 # committed for packages, but should be committed for applications that require a static
 # environment.
-Manifest.toml
+/Manifest.toml
+test/Manifest.toml
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1,0 +1,6 @@
+using SafeTestsets
+
+# Performance and code quality tests
+@safetestset "Aqua tests" begin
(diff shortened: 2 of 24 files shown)


LenkaNovak (approved):
Hi @juliasloan25 , this LGTM. I'm not sure we'll need the experiments directory. What were you thinking would live here?

juliasloan25:
> Hi @juliasloan25 , this LGTM. I'm not sure we'll need the experiments directory. What were you thinking would live here?

I'm not sure, maybe some examples showing how to use the codebase. But the tests should be enough to show the usage of all of the functions. I'll delete the experiments folder for now, and we can always add it back if we need it later.

.gitignore
```diff
@@ -21,4 +21,6 @@ docs/site/
 # It records a fixed state of all packages used by the project. As such, it should not be
 # committed for packages, but should be committed for applications that require a static
 # environment.
-Manifest.toml
+/Manifest.toml
+.dev/Manifest.toml
```
charleskawczynski, line 25: This one might be nice to have?

test/runtests.jl
```diff
@@ -0,0 +1 @@
+
```
charleskawczynski, line 1: Can we add a `@testset` (or better yet, start off with loading SafeTestsets + aqua tests)

PR #9, opened 2023-09-13
Add TimeManager module

### LINKED ISSUE #6: TimeManager module
Create a module for converting, reading, standardizing, etc dates and times. This code was originally developed in ClimaCoupler.jl's TimeManager module, and was incorporated into ClimaLSM.jl's FileReader module, which has less functionality than the coupler's version.

ClimaCoupler.jl TimeManager module: https://github.com/CliMA/ClimaCoupler.jl/blob/main/src/TimeManager.jl
ClimaLSM.jl FileReader module: https://github.com/CliMA/ClimaLSM.jl/blob/main/src/shared_utilities/FileReader.jl

ClimaCoupler.jl TimeManager tests: https://github.com/CliMA/ClimaCoupler.jl/blob/main/test/time_manager_tests.jl
ClimaLSM.jl FileReader tests: https://github.com/CliMA/ClimaLSM.jl/blob/main/test/shared_utilities/file_reader.jl

Closed by #9 

## To Do
- [x] Copy all functions from both repos
- [x] Copy all tests from both repos
- [x] Generalize function interfaces, extend functionality for all users
- [x] Update test usage, make sure tests pass
- [x] Test in user repos
- [x] Add compat entries in project for new deps
- [x] Add docs page

juliasloan25:
## Purpose 
Closes #6

### ClimaCoupler.jl
- `current_date` function removed, should be added to ClimaCoupler.jl Utilities module
- `trigger_callback` interface has changed, now increments cutoff date internally (so the result must get saved wherever the function is called from), and takes in callback function to be called if condition is met
https://github.com/CliMA/ClimaCoupler.jl/issues/423
### ClimaLSM.jl
- no changes :)

----

### DIFF
--- a/docs/src/timemanager.md
+++ b/docs/src/timemanager.md
@@ -1,0 +1,16 @@
+# TimeManager
+
+This module contains functions that handle dates and times
+in simulations. The functions in this module often call
+functions from Julia's [Dates](https://docs.julialang.org/en/v1/stdlib/Dates/) module.
+
+## TimeManager API
+
+```@docs
+ClimaUtilities.TimeManager.to_datetime
+ClimaUtilities.TimeManager.strdate_to_datetime
+ClimaUtilities.TimeManager.datetime_to_strdate
+ClimaUtilities.TimeManager.trigger_callback
+ClimaUtilities.TimeManager.Monthly
+ClimaUtilities.TimeManager.EveryTimestep
+```
--- a/src/TimeManager.jl
+++ b/src/TimeManager.jl
@@ -1,0 +1,97 @@
+"""
+    TimeManager
+
+This module facilitates calendar functions and temporal interpolations
+of data.
+"""
+module TimeManager
+
+import Dates
+import CFTime
+
+export to_datetime,
+    strdate_to_datetime,
+    datetime_to_strdate,
+    trigger_callback,
+    Monthly,
+    EveryTimestep
+
+"""
+    to_datetime(date)
+
+Convert a `DateTime`-like object (e.g. `DateTimeNoLeap`) to a `DateTime`,
+using CFTime.jl. We need this since some data files we use contain
+`DateTimeNoLeap` objects for dates, which can't be used for math with `DateTime`s.
+The `DateTimeNoLeap` type uses the Gregorian calendar without leap years, while
+the `DateTime` type uses Gregorian calendar with leap years.
+
+For consistency, all input data files should have dates converted to `DateTime`
+before being used in a simulation.
+
+# Arguments
+- `date`: `DateTime`-like object to be converted to `DateTime`
+"""
+to_datetime(date) = CFTime.reinterpret(Dates.DateTime, date)
+
+"""
+    strdate_to_datetime(strdate::String)
+
+Convert from String ("YYYYMMDD") to Date format,
+required by the official AMIP input files.
+"""
+strdate_to_datetime(strdate::String) = Dates.DateTime(
+    parse(Int, strdate[1:4]),
+    parse(Int, strdate[5:6]),
+    parse(Int, strdate[7:8]),
+)
+
+"""
+    datetime_to_strdate(datetime::Dates.DateTime)
+
+Convert from DateTime to String ("YYYYMMDD") format.
+"""
+datetime_to_strdate(datetime::Dates.DateTime) =
+    string(lpad(Dates.year(datetime), 4, "0")) *
+    string(string(lpad(Dates.month(datetime), 2, "0"))) *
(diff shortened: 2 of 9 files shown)


src/TimeManager.jl
```diff
@@ -0,0 +1,92 @@
...
+struct Monthly <: AbstractFrequency end
+struct EveryTimestep <: AbstractFrequency end
+
+"""
+    trigger_callback(date_cutoff::Dates.DateTime,
+        date_current::Dates.DateTime,
+        ::Monthly,
+        func::Function,)
+
+If the current date is equal to or later than the cutoff date at time
```
kmdeck, line 66: Is "cutoff" date the clearest name? It seems like the cutoff date is when we want to trigger the callback and save the data. Would `date_nextsave` make sense?

src/TimeManager.jl
```diff
@@ -0,0 +1,92 @@
...
+- `save_freq::AbstractFrequency` frequency with which to trigger callback
+- `func::Function` callback function to be triggered if date is past cutoff
+- `func_args::Tuple` a tuple of arguments to be passed into the callback function
+"""
+function trigger_callback(
+    date_cutoff::Dates.DateTime,
+    date_current::Dates.DateTime,
+    ::Monthly,
+    func::Function,
+    func_args::Tuple,
```
kmdeck, line 82: is any more detail required here? When the callback func is called, what is available to pass in as arguments?

test/timemanager.jl
```diff
@@ -0,0 +1,75 @@
...
+        date_cutoff = date_cutoff_copy = Dates.DateTime(1979, 3, 21)
+        date_current = date_cutoff - Dates.Day(1)
+        date_cutoff = TimeManager.trigger_callback(
+            date_cutoff,
+            date_current,
+            TimeManager.Monthly(),
+            func!,
+            (arg,),
+        )
+        # Test that cutoff date was updated and `func!` got called
```
kmdeck, line 71: Misleading comment? In this case, the func should not be called, and the cutoff date not updated, right?

src/TimeManager.jl
```diff
@@ -0,0 +1,92 @@
...
+    trigger_callback,
+    Monthly,
+    EveryTimestep
+
+"""
+    to_datetime(date)
+
+Convert a DateTime-like object (e.g. DateTimeNoLeap) to a DateTime,
+using CFTime.jl. We need this since some data files we use contain
+DateTimeNoLeap objects for dates, which can't be used for math with DateTimes.
```
LenkaNovak, line 24: Do we need to / should we choose a calendar here?

LenkaNovak (approved):
LGTM, thanks, Julia! Just a few comments.

docs/src/timemanager.md
```diff
@@ -0,0 +1,16 @@
+# TimeManager
+
+This module contains functions that handle dates and times
+in simulations. The functions in this module often call
```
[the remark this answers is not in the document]
juliasloan25: Some of these functions (e.g. `to_datetime`) are used in the bucket model in ClimaLSM, which can be run standalone or in a coupled simulation, so I think it makes sense to keep the comment more general

src/TimeManager.jl
juliasloan25: I added some clarifying comments about `DateTime` vs `DateTimeNoLeap`, the calendars they use, and what we expect when running simulations.

src/TimeManager.jl
juliasloan25: Good point! I'll change it

src/TimeManager.jl
juliasloan25: I just added a comment explaining that they must exactly match the arguments expected by `func`. Besides that, I don't think there are any more restrictions on the arguments to pass in

test/timemanager.jl
juliasloan25: oops, yes

juliasloan25:
@szy21 @glwagner

docs/src/timemanager.md
```diff
@@ -0,0 +1,16 @@
...
+functions from Julia's [Dates](https://docs.julialang.org/en/v1/stdlib/Dates/) module.
+
+## TimeManager API
+
+```@docs
+ClimaCoupler.TimeManager.to_datetime
```
szy21, line 10:
```suggestion
ClimaUtilities.TimeManager.to_datetime
```
(Did you mean this? Same for other APIs)

juliasloan25: oops, good catch!

juliasloan25:
This branch has been tested within ClimaCoupler and works there, so I'm merging it into main

Issue #20, opened 2024-03-26
Drop direct dependency on CFTime


Sbozzolo:
ClimaUtilities depends on `CFTime`, but `CFTime` is used exclusively in one function:
https://github.com/CliMA/ClimaUtilities.jl/blob/8815b952cf27c2cda08204b8df88871b6818fea9/src/TimeManager.jl#L34

I think we can implement that function ourselves:
https://github.com/JuliaGeo/CFTime.jl/blob/d5db3cae2a9f60f4eb980685ada20b3b29ad088e/src/conversions.jl#L81-L86

@juliasloan25 would you like to look into this?

PR #29, opened 2024-04-05
remove CFTime direct dependency

juliasloan25:
## Purpose 
closes https://github.com/CliMA/ClimaUtilities.jl/issues/20

CFTime is still a dependency in test/Project.toml so that we can test the behavior of `reinterpret`

## Content
- [x] add `reinterpret` function to our source code, which was the only function we were using from CFTime
- [x] remove CFTime dependency from top-level src/
- [x] also update docs Manifest, fix a couple of format errors

----

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -5,7 +5,6 @@ version = "0.1.3"
 
 [deps]
 Artifacts = "56f22d72-fd6d-98f1-02f0-08ddc0907c33"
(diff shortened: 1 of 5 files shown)


Sbozzolo:
Could you please add a test to check this?

Sbozzolo (approved):

PR #32, opened 2024-04-11
add default regridder_type

### LINKED ISSUE #23: Automatically select the regridder type if only one is available
Currently, there are two regridders implemnted: `TempestRegridder`, and `InterpolationsRegridder`. At any given time, zero/one/two of them could be available to use (depending on which extension is loaded). 

Having only one regridder available is a common occurence, so we should pick it up automatically, instead of hardcoding that we prefer the `:TempestRegridder`.

### Details
- `:TempestRegridder` is currently hardcoded in:
  - DataHandlingExt.jl DataHandler function
  - SpaceVaryingInputsExt.jl SpaceVaryingInput function
  - TimeVaryingInputsExt.jl TimeVaryingInput function
- TempestRegridderExt.jl requires ClimaCoreTempestRemap.jl to be loaded
- InterpolationsRegridderExt.jl requires Interpolations.jl and ClimaCore.jl to be loaded

### Implementation ideas
- `Base.get_extension(ClimaUtilities, :Extension)` will return the Module Extension if it is loaded, and nothing if it isn't. We can use this to determine which package extensions are loaded.
- pass nothing to DataHandler in default case, choose based on what's loaded in DataHandler only

### Expected behavior
- 4 cases:
  - 2 regridder types loaded (TempestRegridder, InterpolationsRegridder) --> use Interp ?
  - 1 regridder type loaded (TempestRegridder) --> use TR
  - 1 regridder type loaded (InterpolationsRegridder) --> use Interp
  - 0 regridder types loaded --> error

juliasloan25:
## Purpose 
closes #23 

## Content
- Add function `get_regridder_type` to Regridders.jl. This will check which package extensions are loaded and select a regridder to use based on that. By default, interpolation will be used
- Use `get_regridder_type` in DataHandlingExt if regridder type is not supplied by user
- Add tests
  - Note: Tests for 3/4 cases are implemented. I don't think there's a way to clear packages/extensions that are loaded, which we would need to do to check each of the cases of different regridders being available.

----

### DIFF
--- a/src/Regridders.jl
+++ b/src/Regridders.jl
@@ -11,6 +11,8 @@ The key function exposed by `Regridders` is the `regrid` method.
 """
 module Regridders
 
+import ..ClimaUtilities
+
 # When adding a new regridder, you also have to change some functions in the DataHandler
 # module. Find where :TempestRegridder is used.
 abstract type AbstractRegridder end
@@ -20,5 +22,29 @@ function TempestRegridder end
 function InterpolationsRegridder end
 
 function regrid end
+
+"""
+    default_regridder_type()
+
+Return the type of regridder to be used if the user doesn't specify one.
+This function returns the first available regridder in the following order:
+  - InterpolationsRegridder
+  - TempestRegridder
+based on which regridder(s) are currently loaded.
+"""
+function default_regridder_type()
+    # Use InterpolationsRegridder if available
+    if !isnothing(
+        Base.get_extension(ClimaUtilities, :InterpolationsRegridderExt),
+    )
+        regridder_type = :InterpolationsRegridder
+        # If InterpolationsRegridder isn't available, and TempestRegridder is, use TempestRegridder
+    elseif !isnothing(Base.get_extension(ClimaUtilities, :TempestRegridderExt))
+        regridder_type = :TempestRegridder
+    else
+        error("No regridder available")
+    end
+    return regridder_type
+end
 
 end
--- a/docs/src/datahandling.md
+++ b/docs/src/datahandling.md
@@ -43,7 +43,7 @@ the `NCFileReader`.
 
 > This extension is loaded when loading `ClimaCore` and `NCDatasets` are loaded.
 > In addition to this, a `Regridder` is needed (which might require importing
-> additional packages).
+> additional packages) - see [`Regridders`](@ref) for more information.
 
 ## Example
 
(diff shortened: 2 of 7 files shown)


src/Regridders.jl
```diff
@@ -21,4 +23,28 @@ function InterpolationsRegridder end
...
+        Base.get_extension(ClimaUtilities, :InterpolationsRegridderExt),
+    )
+        regridder_type = :InterpolationsRegridder
+        # If InterpolationsRegridder isn't available, and TempestRegridder is, use TempestRegridder
+    elseif !isnothing(Base.get_extension(ClimaUtilities, :TempestRegridderExt))
+        regridder_type = :TempestRegridder
+    else
+        error("No regridder available")
+    end
+    return regridder_type
```
Sbozzolo, line 47: We could be fancy here and do something like (sketch)
```julia
priority_list = Dict(:InterpolationsRegridderExt => :InterpolationsRegridder, ....)
available_regridders = map(ext -> isnothing(Base.get_extension(ClimaUtilities, ext), keys(priority_list))
ret_value = get(priority_list, findfirst(available_regridders), nothing)
isnothing(ret_value) && error("No regridder available")
```

But we don't really need this because we are not going to add more than one other regridder

Sbozzolo (commented):
Can you please also document this behavior in docstrings and docs?

src/Regridders.jl
```diff
@@ -11,6 +11,8 @@ The key function exposed by `Regridders` is the `regrid` method.
 """
 module Regridders
 
+import ClimaUtilities
```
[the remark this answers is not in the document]
juliasloan25: It's needed so we can call `Base.get_extension(ClimaUtilities, _)`. I changed it to `import ..ClimaUtilities` (`import .ClimaUtilities` didn't work)

src/Regridders.jl
juliasloan25: Good point! That would make it easier to change the priority of regridder types too.

I looked into changing it to something like this, but it becomes a little more complicated because `findfirst` returns the index of the first element satisfying our predicate, but we can't index into a `Dict`. We could use `OrderedDict` from OrderedCollections.jl, but I don't want to load another package just for this.

src/Regridders.jl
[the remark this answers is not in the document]
Sbozzolo: If `..ClimaUtilities` works, that would be perfect! :)

src/Regridders.jl
Sbozzolo: Sounds good. No need to overcomplicate this!

Sbozzolo (approved):

PR #34, opened 2024-04-15
outfile_root on all pids

juliasloan25:
## Purpose 
Create `outfile_root` on all pids, so we can use TempestRegridder on MPI. Also displays `context` in the data_handling test, which will be helpful to ensure we're running on the architecture we expect.


Sbozzolo (approved):
Let's merge this after #33

PR #35, opened 2024-04-16
add LRU cache

### LINKED ISSUE #21: Control/clean caches
Currently, the caches in `FileReaders` and `DataHandling` are unlimited. Every time a new field is added, the cache grows, and there is no direct way to purge them. Obviously, this is not a great idea. 

The way forward:
- Add a function to manually clean the cache
- Implement a LRU (least recently used) cache with a fixed maximum size

juliasloan25:
## Purpose 
closes #21

## Content
- [x] add Cache module containing an LRUCache struct, `get!` function, and `enforce_size!` function
- [x] add tests for Cache module
- [x] use LRUCache in FileReader and DataHandling

### DIFF
--- a/docs/src/datahandling.md
+++ b/docs/src/datahandling.md
@@ -37,6 +37,9 @@ provided to the `DataHandler`.
 
 The `DataHandler` has a caching mechanism in place: once a field is read and
 regridded, it is stored in the local cache to be used again (if needed).
+This is a least-recently-used (LRU) cache implemented in `DataStructures`,
+which removes the least-recently-used data when its maximum size is reached.
+The default maximum size is 128.
 
 While the reading backend could be generic, at the moment, this module uses only
 the `NCFileReader`.
--- a/docs/src/datastructures.md
+++ b/docs/src/datastructures.md
@@ -1,0 +1,63 @@
+# `DataStructures`
+
+The `DataStructures` module implements helpful data structures to be used by
+other ClimaUtilities.jl modules or external packages.
+
+## `LRUCache`
+
+`DataStructures` implements an `LRUCache`, which is used in both `DataHandlingExt`
+and `NCFileReaderExt`. `LRUCache` can be used to store pieces of information that
+need to be accessed multiple times and may be expensive to compute, such as
+regridded fields or loaded files. Instead of recomputing the values each time
+they're needed, the previously-computed information is saved in the cache and
+retrieved directly from it. To prevent the cache from growing so large that it
+takes up significant memory, the least-recently-used (LRU) scheme maintains
+a maximum cache size: every time adding an element to the cache would
+lead its size to grow larger thant the maximum allowed, the element that was
+accessed the least recently is deleted first.
+
+> Note: all the methods supported by dictionaries are currently implemented.
+If you need one that is not implemented, please open an issue or a pull request.
+
+### Example
+In many ways, `LRUCache`s behave like Julia dictionaries. To use one, we
+first need to initialize the empty cache specifying types of keys and values
+and optionally the maximum allowed size:
+
+```julia
+import ClimaUtilities.DataStructures
+cache = DataStructures.LRUCache{Int, Int}(; max_size = 128)
+```
+
+Once we have the `cache`, we can access and insert elements with `get!`.
+`get!` retrieves the value associated to the key if avaialble, otherwise
+it inserts a new key with a provided default. The default can be passed as
+third argument or can be the return statement of a function provided as
+first (as typically done in `do` blocks). Continuing the example above
+
+```julia
+println("The value associated to 1 is: ", get!(cache, 1, 10))
+other_value = get!(cache, 2) do
+    return 20
+end
+```
+
+This cache can be used to implement some recursive function more efficiently,
+as in the famous case of the factorial:
+```julia
+function factorial(n)
+    n in (0, 1) && return 1
+    return n * get!(cache, n - 1, factorial(n - 1))
+end
+# The first time we call factorial it will compute everything
+@time factorial(10)
+# The second time it will only compute the last element
+@time factorial(11)
+```
+
+## API
+
+```@docs
+ClimaUtilities.DataStructures.LRUCache
+Base.get!
+```
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -70,5 +70,11 @@ The `DataHandling` module bundles a `Regridder` and a `FileReader` together to
 serve regridded fields at a given time upon request. The main interface for
 `DataHandling` is `regridded_snapshot(data_handler, date)`, a function that
 returns a `Field` with data read from file for the given `date`. The
-`DataHandler` maintains a cache of regridded fields to amortize the cost of
-(expensive) regridding operations.
+`DataHandler` maintains an least-recently-used (LRU) cache of regridded fields
+to amortize the cost of (expensive) regridding operations.
+
+### `DataStructures`
+
+The `DataStructures` module implements helpful data structures to be used by
+other ClimaUtilities.jl modules or external packages. Currently it contains an
+LRU cache that is used in `DataHandlingExt` and `NCFileReaderExt`.
--- a/src/DataStructures.jl
+++ b/src/DataStructures.jl
@@ -1,0 +1,102 @@
+"""
+    DataStructures
+
+The `DataStructures` module implements useful data structures.
+
+Currently, we have implemented a Least Recently Used (LRU) cache. The cache is
+implemented as a dictionary with a maximum size. When the cache is full, the
+least recently used item is removed to make space for the new item.
+
+Within ClimaUtilities, the LRU cache is used by the `FileReaders` module to
+store files that are currently open, and by the `DataHandler` module to store
+regridded fields.
+"""
+module DataStructures
+
+import Base: Callable
+
+export LRUCache
+
+struct LRUCache{K, V}
+    """The cache itself, containing key-value pairs of information."""
+    cache::Dict{K, V}
+
+    """The maximum number of key-value pairs in the cache."""
+    max_size::Int
+
+    """A list of keys, ordered by their last access time, which serves as a
+    priority queue for the cache. Note that another data structure could be
+    more efficient here, but a built-in vector is sufficient given that
+    the cache is not expected to be very large."""
+    priority::Vector{K}
+end
+
+"""
+    LRUCache{K, V}(; max_size::Int = 128) where {K, V}
+
+Construct an empty `LRUCache` with a maximum size of `max_size`.
+"""
+function LRUCache{K, V}(; max_size::Int = 128) where {K, V}
+    return LRUCache{K, V}(Dict{K, V}(), max_size, Vector{K}())
+end
+
+"""
+    Base.get!(cache::LRUCache{K, V}, key::K, default::V) where {K, V}
+
+Get the value associated with `key` in `cache`. If the key is not in the
+cache, add it with the value `default`. In any case, update the key's priority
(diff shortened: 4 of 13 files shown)


src/Cache.jl
```diff
@@ -0,0 +1,95 @@
...
+
+The cache is used by the `FileReaders` module to store files that are currently
+open, and by the `DataHandler` module to store regridded fields.
+"""
+
+module Cache
+
+import Base: Callable
+
+export LRUCache, get!, enforce_size!
```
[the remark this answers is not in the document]
Sbozzolo: You also don't need to export `get!` since you are adding a method to `Base.get!`

src/Cache.jl
```diff
@@ -0,0 +1,95 @@
...
+
+struct LRUCache{K, V}
+    cache::Dict{K, V}
+    max_size::Int
+    priority::Vector{K}
```
Sbozzolo, line 23: Can you add a comment on the choice of Vector for your `priority`? `Vector` is not the best datastructure to use here for its asymptotic scaling, but it is fine for what we want to do here given that it is built in and the size of the caches we are working with is not massive.

src/Cache.jl
```diff
@@ -0,0 +1,95 @@
...
+    if haskey(cache.cache, key)
+        # Remove key from priority queue so we can update it
+        filter!(k -> k != key, cache.priority)
+    else
+        cache.cache[key] = default
+    end
+    # Update priority of key
+    push!(cache.priority, key)
+    enforce_size!(cache)
+    return cache.cache[key]
```
Sbozzolo, line 52: Maybe something like this:
```suggestion
update_priority!(cache, key)
enforce_size!(cache)
return get!(cache.cache, key, default)
```
With 
```julia
function update_priority(cache, key)
filter!(k -> k != key, cache.priority)
push!(cache.priority, key)
```
(In either cases we want the new key to be the last entry). 

And `enforce_size` that enforces `max_size - 1`

src/Cache.jl
```diff
@@ -0,0 +1,95 @@
...
+    if haskey(cache.cache, key)
+        # Remove key from priority queue so we can update it
+        filter!(k -> k != key, cache.priority)
+    else
+        cache.cache[key] = default()
+    end
+    # Update priority of key
+    push!(cache.priority, key)
+    enforce_size!(cache)
+    return cache.cache[key]
```
Sbozzolo, line 78:
```suggestion
update_priority!(cache, key)
enforce_size!(cache)
return get!(default, cache.cache, key)
```

src/Cache.jl
```diff
@@ -0,0 +1,95 @@
...
+
+Remove the least recently used items from `cache` until its size is
+less than or equal to `cache.max_size`.
+"""
+function enforce_size!(cache::LRUCache{K, V}) where {K, V}
+    while length(cache.priority) > cache.max_size
+        key = popfirst!(cache.priority)
+        delete!(cache.cache, key)
+    end
+end
```
Sbozzolo, line 92: I think we should be able to assume that the size can only be one unit larger than the maximum. In this case, we can remove the `while`. If you want to support arbitrary sized, iterative `popfirst!` does not scale very well, it would be better to remove the first N elements in one go.

Sbozzolo (commented):
I think that this a useful data structure, so I would say that we should expose this to other packages that want to use it (eg Coupler, Atmos, ..).

We could rename the `Cache` module to `DataStructures` and add some documentation to the docs.

What do you think?

src/Cache.jl
Sbozzolo: Since the code is identical, you could use dispatch to merge the two `get!` into one.

Sbozzolo:
There's also some comments/TODO sprinkled around the code that need to be updated.

src/Cache.jl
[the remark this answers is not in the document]
juliasloan25: is the `_function` syntax for internal functions?

juliasloan25:
> I think that this a useful data structure, so I would say that we should expose this to other packages that want to use it (eg Coupler, Atmos, ..).
> 
> We could rename the `Cache` module to `DataStructures` and add some documentation to the docs.
> 
> What do you think?

That sounds good to me! I'll rename it and update the docs

src/Cache.jl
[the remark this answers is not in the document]
Sbozzolo: > is the `_function` syntax for internal functions?

It is a possible convention, mostly used in Python. SciML uses two underscores to denote private functions (https://docs.sciml.ai/SciMLStyle/stable/). CliMA is not consistent in what we do (but we mostly don't use this convention). I think it is a useful convention (it tells readers that they shouldn't use that function), but it by no means universally agreed upon.

src/Cache.jl
Sbozzolo: `cache.cache` is a Dictionary so this is not recursive. It is implementing our `get!` in terms of Julia's `get!`, which

src/Cache.jl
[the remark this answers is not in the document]
juliasloan25: that makes sense, thank you!

src/Cache.jl
juliasloan25: Ahh, I missed that it was calling `get!` on `cache.cache`

docs/src/datahandling.md
```diff
@@ -37,6 +37,9 @@ provided to the `DataHandler`.
 
 The `DataHandler` has a caching mechanism in place: once a field is read and
 regridded, it is stored in the local cache to be used again (if needed).
+This is an LRU cache implemented in `DataStructures`, which removes the
```
Sbozzolo, line 40:
```suggestion
This is an least recently used cache implemented in `DataStructures`, which removes the
```

docs/src/datastructures.md
```diff
@@ -0,0 +1,28 @@
+# `DataStructures`
+
+The `DataStructures` module implements helpful data structures to be used by
+other ClimaUtilities.jl modules or external packages.
+
+`DataStructures` implements an `LRUCache`, which is used in both `DataHandlingExt`
```
Sbozzolo, line 6: Can you describe what problem `LRUCache` is solving? When would someone use it?

docs/src/datastructures.md
```diff
@@ -0,0 +1,28 @@
...
+and `NCFileReaderExt`. The `LRUCache` contains the following fields:
+- `cache`: A dictionary containing key-value pairs of information we've
+    recently accessed.
+- `max_size`: The maximum number of key-value pairs that can be stored in `cache`.
+- `priority`: A vector of keys that is sorted based on their priority
+    (i.e. when they were last accessed/used). The least-recently-used entry is
+    at the front, and will be removed from both `priority` and `cache` when the
+    maximum size of the cache is reached.
```
Sbozzolo, line 14: I think that this is a technical detail and users do not have to know how `LRUCache` is implemented to use it.

docs/src/index.md
```diff
@@ -70,5 +70,11 @@ The `DataHandling` module bundles a `Regridder` and a `FileReader` together to
...
 returns a `Field` with data read from file for the given `date`. The
-`DataHandler` maintains a cache of regridded fields to amortize the cost of
+`DataHandler` maintains an LRU cache of regridded fields to amortize the cost of
 (expensive) regridding operations.
+
+### `DataStructures`
+
+The `DataStructures` module implements helpful data structures to be used by
+other ClimaUtilities.jl modules or external packages. Currently it contains an
+LRU cache that is used in `DataHandlingExt` and `NCFileReaderExt`.
```
Sbozzolo, line 80:
```suggestion
least recently used cache that is used in `DataHandlingExt` and `NCFileReaderExt`.
```

src/DataStructures.jl
```diff
@@ -0,0 +1,102 @@
...
+end
+
+"""
+    LRUCache{K, V}(max_size::Int) where {K, V}
+
+Construct an empty `LRUCache` with a maximum size of `max_size`.
+"""
+function LRUCache{K, V}(max_size::Int) where {K, V}
+    return LRUCache{K, V}(Dict{K, V}(), max_size, Vector{K}())
+end
```
Sbozzolo, line 41: I'd make `max_size` a keyword argument with default value 128 (`max_size` is a "parameter").

docs/src/datastructures.md
juliasloan25: I added some more info. Let me know if you think it should be tweaked!

docs/src/datastructures.md
```diff
@@ -0,0 +1,38 @@
...
+## `LRUCache`
+
+`DataStructures` implements an `LRUCache`, which is used in both `DataHandlingExt`
+and `NCFileReaderExt`. `LRUCache` can be used to store pieces of information that
+need to be accessed multiple times and may be expensive to compute, such as
+regridded fields or loaded files. Instead of recomputing the values each time
+they're needed, the previously-computed information is saved in the cache and
+retrieved directly from it. To prevent the cache from growing so large that it
+takes up significant memory, the least-recently-used (LRU) scheme maintains
+a maximum cache size.
```
Sbozzolo, line 15:
```suggestion
retrieved directly from it. To prevent the cache from growing so large that it
takes up significant memory, the least-recently-used (LRU) scheme maintains
a maximum cache size: every time adding an element to the cache would 
lead its size to grow larger thant the maximum allowed, the element that was
accessed the least recently is deleted first. 
```

docs/src/datastructures.md
```diff
@@ -0,0 +1,38 @@
...
+
+The main function used to interact with an `LRUCache` is `get!`, which has two methods:
+- `get!(cache::LRUCache{K, V}, key::K, default::V)`: This method will retrieve
+    the value stored in the cache at `key` if it exists. Otherwise, it will add
+    an entry at `key` having value `default`. In either case, the priority of this
+    entry will be updated.
+- `get!(default::Callable, cache::LRUCache{K, V}, key::K)`: This method will retrieve
+    the value stored in the cache at `key` if it exists. Otherwise, it will add
+    an entry at `key` having value `default()`. This can be used with the `do`-block
+    syntax. As before, in either case, the priority of this entry will be updated.
```
Sbozzolo, line 38: We can move this to the module itself and have a more user centric documentation.

````suggestion
In many ways, `LRUCache`s behave like Julia dictionaries. To use one, we first need to initialize the empty cache specifying types of keys and values and optionally the maximum allowed size:
```julia
import ClimaUtilities.DataStructures
cache = DataStructures.LRUCache{Int, Int}(; max_size = 128)
```
Once we have the `cache`, we can access and insert elements with `get!`. `get!` retrieves the value associated to the key if avaialble, otherwise it inserts a new key with a provided default. The default can be passed as third argument or can be the return statement of a function provided as first (as typically done in `do` blocks). Continuing the example above
```julia
println("The value associated to 1 is: ", get!(cache, 1, 10))
other_value = get!(cache, 2) do
    return 20
end
```
This cache can be used to implement some recursive function more efficiently, as in the famous case of the factorial:
```julia
function factorial(n)
    n in (0, 1) && return 1
    return n * get!(cache, n - 1, factorial(n - 1)
end
# The first time we call factorial it will compute everything
@time factorial(10)
# The second time it will only compute the last element
@time factorial(11)
````

> Note: all the methods supported by dictionaries are currently implemented. If you need one that is not implemented, please open an issue or a pull request.
```

docs/src/datastructures.md
Sbozzolo: We should probably have a APIs section to follow the other documentation pages.

Sbozzolo:
Also, please add an entry to the NEWS file :)

juliasloan25:
@Sbozzolo Thanks for all the feedback, hopefully everything looks good now :)

PR #36, opened 2024-04-17
bump patch version

juliasloan25:
## Purpose 
To be merged after #33, so downstream packages can use these fixes


Sbozzolo (approved):
Let's merge this after the cache is merged.

PR #40, opened 2024-04-18
Add support for passing arguments down to the regridder and file reader

### LINKED ISSUE #22: Pass additional arguments to the regridder from higher level constructors
It is not possible to specify additional arguments to the regridder from higher level interfaces. For example, it is not possible to change the `mono` keyword in the TempestRegridder.

We should expose additional (keyword/) arguments in higher level interfaces.

Sbozzolo:
This PR exposes some constructor options to the higher level abstractions.

Closes #22 

cc: @kmdeck

### DIFF
--- a/docs/src/regridders.md
+++ b/docs/src/regridders.md
@@ -63,13 +63,27 @@ along each direction) and returns a `ClimaCore` `Field` defined on the
 Currently, `InterpolationsRegridder` only supports spherical shells and extruded
 spherical shells (but it could be easily extended to other domains).
 
-> Note: it is easy to change the spatial interpolation type and extrapolation
-> conditions, if needed.
+> Note: it is easy to change the spatial interpolation type if needed.
 
 `InterpolationsRegridder` are created once, they are tied to a `target_space`,
 but can be used with any input data. With MPI runs, every process computes the
 interpolating function. This is always done on the CPU and moved to GPU for
 accelerated runs.
+
+By default, `InterpolationsRegridder` assumes you are interpolating on a globe
+and the default extrapolation boundary conditions are: periodic (along
+longitudes), copy (along latitude), and throwing an error (along z). These can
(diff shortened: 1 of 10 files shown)


ext/DataHandlingExt.jl
```diff
@@ -133,25 +141,28 @@ function DataHandling.DataHandler(
         regridder_type
 
     # File reader, deals with ingesting data, possibly buffered/cached
-    file_reader = NCFileReader(file_path, varname)
+    file_reader = NCFileReader(file_path, varname; file_reader_kwargs...)
```
juliasloan25, line 144: It would be good to add some verification that the kwargs are valid inputs to the constructors we're passing them to (for both file reader and regridder)

test/data_handling.jl
```diff
@@ -59,6 +59,32 @@ ClimaComms.init(context)
             close(data_handler)
         end
     end
+    # Test passing arguments down
```
juliasloan25, line 62: Maybe we should test for the file reading case as well as regridding

juliasloan25 (commented):
Looks good! I just left a few suggestions

docs/src/regridders.md
```diff
@@ -63,14 +63,28 @@ along each direction) and returns a `ClimaCore` `Field` defined on the
...
 but can be used with any input data. With MPI runs, every process computes the
 interpolating function. This is always done on the CPU and moved to GPU for
 accelerated runs.
 
+By default, `InterpolationsRegridder` assumes you are interpolating on a globe
+and the default extrapolation boundary conditions are: periodic (along
+longitudes), copy (along latitude), and throwing an error (along z). These can
+be changed by passing the `extrapolation_bc` to the constructor of the regridder.
+
+## FAQ: How do I enable liner extrapolation in z?
```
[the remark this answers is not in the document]
Sbozzolo: Thanks!

juliasloan25 (approved):

PR #42, opened 2024-04-20
Add README

### LINKED ISSUE #43: The readme for ClimaUtilities is incomplete

Sbozzolo:
Closes #43

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,8 +3,21 @@
 ClimaUtilities.jl
 </h1>
 
-Shared utilities for packages within the CliMA project.
+Shared utilities for packages within the CliMA project. Please, check out the
+[documentation](https://clima.github.io/ClimaUtilities.jl/dev/).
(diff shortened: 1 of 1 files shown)


juliasloan25 (commented):
This looks good! Maybe we can also add `DataStructures` now?

Also, it would be nice to add a line explaining that the origin of most of this code is ClimaCoupler, to give credit to its developers - unless you feel like the design has become different enough that we've diverged from what was implemented there originally

Sbozzolo:
I think that saying that origins of this code are from ClimaCoupler is no longer accurate. The only components that are from ClimaCoupler are `TimeManager` and `TempestRegridder`, the `Input`s are partially from `Land` (and Kat should be credited for the work on the SpatialInputs), everything else is new. 

I will update the readme with the DataStructures and pointing that out.

PR #44, opened 2024-04-22
Fix MPI test 

Sbozzolo:
https://buildkite.com/clima/climautilities-ci/builds/127#018efc0a-213e-406b-a629-e676dd2657d1

The issue is that some MPI processes could run `rm` on a filesystem without the socrates file. By forcing it, all the MPI processes will be able to run `rm` without complaining if no file is there.


juliasloan25 (approved):
It looks like buildkite precompilation is failing, but after that's sorted I think this is good to go

Sbozzolo:
> It looks like buildkite precompilation is failing, but after that's sorted I think this is good to go

I think it is because of a corrupted depot due to the filesystem issues. I cleaned the depot and now it is running.

PR #46, opened 2024-04-22
Add downstream tests

### LINKED ISSUE #45: Add downstream tests
As this repo gets used more and more by downstream packages, the risk of accidentally breaking things increases. I think it would be good to add new CI matrix where we run the test suite for ClimaLand, ClimaAtmos (and ClimaCoupler in the future).

Sbozzolo:
Closes #45

### DIFF
--- a/.github/workflows/downstream.yml
+++ b/.github/workflows/downstream.yml
@@ -1,0 +1,37 @@
+name: Downstream
+on:
+  push:
+    branches:
+      - main
+    tags: '*'
+  pull_request:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  test:
+    name: downstream ${{ matrix.package }}
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        package:
+          - 'ClimaAtmos.jl'
+          - 'ClimaLand.jl'
+    steps:
+      - uses: actions/checkout@v4
+      - uses: julia-actions/setup-julia@v1
+        with:
+          version: '1.10'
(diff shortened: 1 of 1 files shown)


Sbozzolo:
The test that is failing in ClimaLand doesn't have anything to do with ClimaUtilities...

.github/workflows/downstream.yml
```diff
@@ -0,0 +1,38 @@
...
+      - uses: julia-actions/cache@v1
+      - uses: julia-actions/julia-buildpkg@v1
+      - uses: actions/checkout@v4
+        with:
+          repository: 'CliMA/${{ matrix.package }}'
+          path: ${{ matrix.package }}
+      - run: |
+          julia --project=${{ matrix.package }} -e 'using Pkg; Pkg.instantiate()'
+          julia --project=${{ matrix.package }} -e 'using Pkg; Pkg.develop(; path = ".")'
+          julia --project=${{ matrix.package }} -e 'using Pkg; Pkg.test()'
```
juliasloan25, line 38: can we add `--color=yes` here please?

Sbozzolo: Added

juliasloan25:
It would be nice to merge this - maybe we can just increase the tolerance for the land test

Sbozzolo:
> It would be nice to merge this - maybe we can just increase the tolerance for the land test

I don't understand how eps(FT) could be 1e-7. 

It is also quite annoying to debug this.

The easiest way out is to increase the tolerance by one order of magntiude again, but this is not very satisfying solution

Sbozzolo:
@juliasloan25 the issue with ClimaLand got fixed (it was `arch = x86` in the action), so I think we can merge this!

juliasloan25 (approved):

PR #47, opened 2024-04-23
Bump to 0.1.5

Sbozzolo:
ClimaLand wants passing down arguments for the global runs.


juliasloan25:
Why did you delete `docs/Manifest.toml`? It's checked in in a lot of our repos, including ClimaCore, ClimaAtmos, and ClimaLand

Sbozzolo:
> Why did you delete `docs/Manifest.toml`? It's checked in in a lot of our repos, including ClimaCore, ClimaAtmos, and ClimaLand

Maintaining `Manifest.toml` files is a pain with only one upside: reproducibility. While packages like ClimaLand execute some code in the documentation (eg for the tutorials), the documentation for ClimaUtilities is just text and it should be rendered approximately in the same way regardless of the versions of the underlying packages. 

I think that the default should be to not check in the `docs/Manifest.toml`, and add it only if really needed (so, only in ClimaLand).As an example, Documenter.jl does not have a Manifest.toml: https://github.com/JuliaDocs/Documenter.jl/tree/master/docs

juliasloan25:
> > Why did you delete `docs/Manifest.toml`? It's checked in in a lot of our repos, including ClimaCore, ClimaAtmos, and ClimaLand
> 
> Maintaining `Manifest.toml` files is a pain with only one upside: reproducibility. While packages like ClimaLand execute some code in the documentation (eg for the tutorials), the documentation for ClimaUtilities is just text and it should be rendered approximately in the same way regardless of the versions of the underlying packages.
> 
> I think that the default should be to not check in the `docs/Manifest.toml`, and add it only if really needed (so, only in ClimaLand).As an example, Documenter.jl does not have a Manifest.toml: https://github.com/JuliaDocs/Documenter.jl/tree/master/docs

Okay, that sounds reasonable to me

juliasloan25 (approved):

PR #50, opened 2024-04-25
Do not import CUDA from ClimaCore and bump to 0.1.6

### LINKED ISSUE #48: Fix imports
We probably shouldn't import CUDA/adapt in this way: `import ClimaCore.Fields: Adapt, CUDA`

Found in this build (when moving CUDA to an extension in ClimaCore): https://buildkite.com/clima/climaatmos-ci/builds/18491#018f108f-a28b-464d-9008-d81246d5e9a3

Sbozzolo:
Closes #48 

Ideally, one would probably want to consolidate the two calls and do not call CUDA directly. 

Let's merge this after ClimaComms splits CUDA off too.


Project.toml
```diff
@@ -28,7 +28,7 @@ TimeVaryingInputsExt = ["ClimaCore", "NCDatasets"]
 [compat]
 Adapt = "3, 4"
 Artifacts = "1"
-ClimaComms = "0.5"
+ClimaComms = "0.5, 0.6"
 ClimaCore = "0.11, 0.12, 0.13"
```
juliasloan25, line 32: I think this compat will need to include ClimaCore 0.14 so we're actually compatible with it

ext/TimeVaryingInputs0DExt.jl
```diff
@@ -84,12 +84,27 @@ function TimeVaryingInputs.evaluate!(
 )
     time in itp || error("TimeVaryingInput does not cover time $time")
     if ClimaComms.device(itp.context) isa ClimaComms.CUDADevice
-        CUDA.@cuda TimeVaryingInputs.evaluate!(
-            parent(destination),
-            itp,
-            time,
-            itp.method,
-        )
+        if pkgversion(ClimaComms) >= v"0.6"
```
juliasloan25, line 87: Could you add a comment explaining that starting in ClimaComms v0.6, CUDA isn't automatically loaded via ClimaComms?

src/OutputPathGenerator.jl
```diff
@@ -180,14 +180,21 @@ function generate_output_path(::ActiveLinkStyle, output_path; context = nothing)
...
+            filter(x -> !isnothing(match(name_rx, x)), readdir(output_path))
+        if length(existing_outputs) > 0
+            @warn "$output_path already contains some output data, but no active link"
+            latest_output = first(sort(existing_outputs, rev = true))
+            counter_str = match(name_rx, latest_output)
+            counter = parse(Int, counter_str[1])
+            next_counter = counter + 1
+            @warn "Restarting counter from $next_counter"
+        else
+            # This is our fist counter
```
juliasloan25, line 195: typo: fist -> first ?

src/OutputPathGenerator.jl
```diff
@@ -180,14 +180,21 @@ function generate_output_path(::ActiveLinkStyle, output_path; context = nothing)
             )
         end
     else
-        # The link does not exist, but maybe there are already output folders
-        if any(map(x -> !isnothing(match(name_rx, x)), readdir(output_path)))
-            error(
-                "$output_path already contains some output data, but no active link",
-            )
+        # The link does not exist, but maybe there are already output folders.
+        # We can try to guess what was the last one
```
juliasloan25, line 184: could you explain briefly in a comment how you try to find the last one?

NEWS.md
```diff
@@ -1,9 +1,12 @@
 ClimaUtilities.jl Release Notes
 ===============================
 
-v0.1.5
+v0.1.6
 -------
+- Fix compatibility with ClimaCore 0.14. PR [#50](https://github.com/CliMA/ClimaUtilities.jl/pull/50)
```
juliasloan25, line 6: maybe also note that the OutputPathGenerator will also try to find the last output path instead of erroring now

juliasloan25 (commented):
Thanks Gabriele! I left some suggestions, mostly to add more comments. I also think it's a good idea to wait for the ClimaCore v0.14 release (today, hopefully), so we can include that in our compat

Project.toml
juliasloan25: Actually maybe this depends on if ClimaCore v0.14.0 is released before or after it's updated for ClimaComms v0.6 https://github.com/CliMA/ClimaCore.jl/pull/1697

Sbozzolo: Technically, moving to ClimaComms 0.6 is ClimaCore could be considered breaking because CUDA is no longer available by default. So, I think it would be better to have it with 0.14

juliasloan25 (approved):

PR #57, opened 2024-05-12
Do not allow empty output_path

### LINKED ISSUE #56: `generate_output_path` fails with "writing to read-only" when passed empty `output_path`
Failure example [here](https://github.com/CliMA/ClimaAtmos.jl/actions/runs/9020764355/job/24786627579?pr=2994#step:6:1489).

Sbozzolo:
This PR explicitely disallows empty `output_path`s in `generate_output_path` for the `ActiveLinkStyle`. Previously, this input was not handled correctly: when `""` was passed, a slash would be added, so that `generate_output_path` tried writing to `/output_0000`, which is read-only. Now, we inform the user that their input is invalid.

Fixes #56

### DIFF
--- a/src/OutputPathGenerator.jl
+++ b/src/OutputPathGenerator.jl
@@ -110,6 +110,7 @@ function generate_output_path(
     context = nothing,
     style = ActiveLinkStyle(),
 )
+    output_path == "" && error("output_path cannot be empty")
     return generate_output_path(style, output_path; context)
 end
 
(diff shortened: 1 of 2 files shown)


charleskawczynski (commented):
Looks good, should we also apply this to the `RemovePreexistingStyle` method for consistency?

Sbozzolo:
> Looks good, should we also apply this to the `RemovePreexistingStyle` method for consistency?

Right. I just added to the main entry point so that it is implemented for all the styles.

charleskawczynski (approved):

PR #58, opened 2024-05-12
Use NTFS junction points when symlinks cannot be created on Windows

### LINKED ISSUE #52: Symlinks require admin privileges on windows
```
ERROR: LoadError: IOError: On Windows, creating symlinks requires Administrator privileges.
symlink("output_0000", "output\\10secs_3600secs_z_elem_20\\output_active"): operation not permitted (EPERM)
Stacktrace:
  [1] uv_error
    @ .\libuv.jl:100 [inlined]
  [2] symlink(target::String, link::String; dir_target::Bool)
    @ Base.Filesystem .\file.jl:1112
  [3] symlink
    @ .\file.jl:1074 [inlined]
  [4] generate_output_path(::ClimaUtilities.OutputPathGenerator.ActiveLinkStyle, output_path::String; context::ClimaComms.SingletonCommsContext{…})
```

Sbozzolo:
Windows does not always allow the creation of symbolic links by unprivileged users. This depends on the version of Windows, but also some of its settings. When the creation of symbolic links is not possible, `OutputPathGenerator` will create NTFS junction points instead. Junction points are similar to symbolic links, with the main difference that they
have to refer to directories and they have to be absolute paths. As a result, on systems that do not allow unprivileged users to create symbolic links, moving the base output folder results in breaking the `output_active` link.

Note, I have no way to test if this code is valid besides relying on the CI runners. Additional validation would be helpful.

As a reference, here is the docstring for `symlink`:
```julia
  symlink(target::AbstractString, link::AbstractString; dir_target = false)

  Creates a symbolic link to target with the name link.

  On Windows, symlinks must be explicitly declared as referring to a directory
  or not. If target already exists, by default the type of link will be auto-
  detected, however if target does not exist, this function defaults to
  creating a file symlink unless dir_target is set to true. Note that if the
  user sets dir_target but target exists and is a file, a directory symlink
  will still be created, but dereferencing the symlink will fail, just as if
  the user creates a file symlink (by calling symlink() with dir_target set to
  false before the directory is created) and tries to dereference it to a
  directory.

  Additionally, there are two methods of making a link on Windows; symbolic
  links and junction points. Junction points are slightly more efficient, but
  do not support relative paths, so if a relative directory symlink is
  requested (as denoted by isabspath(target) returning false) a symlink will
  be used, else a junction point will be used. Best practice for creating
  symlinks on Windows is to create them only after the files/directories they
  reference are already created.

  See also: hardlink.

  │ Note
  │
  │  This function raises an error under operating systems that do not
  │  support soft symbolic links, such as Windows XP.

  │ Julia 1.6
  │
  │  The dir_target keyword argument was added in Julia 1.6. Prior to
  │  this, symlinks to nonexistent paths on windows would always be
  │  file symlinks, and relative symlinks to directories were not
  │  supported.
```

Fixes #52

### DIFF
--- a/src/OutputPathGenerator.jl
+++ b/src/OutputPathGenerator.jl
@@ -77,6 +77,16 @@ Let us assume that `output_path = dormouse`.
 - `dormouse` exists and does not contain a `output_active`, `ActiveLinkStyle` will check if
   any `dormouse/output_XXXX` exists. If not, it creates `dormouse/output_0000` and a link
   `dormouse/output_active` that points to this directory.
+
+## A note for Windows users
+
+Windows does not always allow the creation of symbolic links by unprivileged users. This
+depends on the version of Windows, but also some of its settings. When the creation of
+symbolic links is not possible, `OutputPathGenerator` will create NTFS junction points
+instead. Junction points are similar to symbolic links, with the main difference that they
+have to refer to directories and they have to be absolute paths. As a result, on systems
+that do not allow unprivileged users to create symbolic links, moving the base output folder
+results in breaking the `output_active` link.
 """
 struct ActiveLinkStyle end
 
@@ -209,8 +219,23 @@ function generate_output_path(::ActiveLinkStyle, output_path; context = nothing)
     # Make new folder
     if root_or_singleton(context)
         mkpath(new_output_folder)
-        # In Windows, symlinks must be explicitly declared as referring to a directory or not
-        symlink("output_$next_counter_str", active_link, dir_target = true)
+        # On Windows, creating symlinks might require admin privileges. This depends on the
+        # version of Windows and some of its settings (e.g., if "Developer Mode" is enabled).
+        # So, we first try creating a symlink. If this fails, we resort to creating a NTFS
+        # junction. This is almost the same as a symlink, except it requires absolute paths.
+        # In general, relative paths would be preferable because they make the output
+        # completely relocatable (whereas absolute paths are not).
+        try
+            symlink("output_$next_counter_str", active_link, dir_target = true)
+        catch e
+            if e isa Base.IOError && Base.uverrorname(e.code) == "EPERM"
+                active_link = abspath(active_link)
+                dest_active_link = abspath("output_$next_counter_str")
+                symlink(dest_active_link, active_link, dir_target = true)
+            else
+                rethrow(e)
+            end
+        end
     end
     maybe_wait_filesystem(context, new_output_folder)
     return active_link
--- a/docs/src/outputpathgenerator.md
+++ b/docs/src/outputpathgenerator.md
@@ -53,6 +53,14 @@ Let's assume your `output_path` is set to `data`.
   none are found, it creates `data/output_0000` and a link `data/output_active`
   pointing to it.
 
+#### A note for Windows users
+
+Windows does not always allow the creation of symbolic links by unprivileged
+users, so some details about links might be slightly different depending on your
+system. If you are using Windows, please have a look at docstring on the
+`ActiveLinkStyle` to learn more about possible differences.
+
+
 ## API
 
 ```@docs


Sbozzolo:
> I would argue that `ActiveLinkStyle` is no longer a good name if it's not platform agnostic. It seems like an awkward restriction for users to specify full paths (and inconsistent if it's only on windows).

The use of junction points is internal in julia (they are still created by the `symlink` function). Junction points are symlinks with two extra restrictions: they require full paths and only work with directories. Users of the `OutputPathGenerator` do not have to specify full paths: the PR prepends them automatically when they are needed. 

For the point of view of users, the only difference between OSs is that if your OS doesn't support actual symlinks, you won't be able to move the `base_output_directory` (without breaking the link). 

> What's the benefit of using Symlinks? I'm wondering what a solution to not overwriting data might look like without symlinks.

This is the idea I came up with trying to find a solution that 
- doed not touch existing data
- provides a fixed and predicatable output path for the newest data
- has nearly zero performance cost

The way I came up with to do this without symlinks involved moving data (e.g., the output is always is a new temporarly folder, but then we "file" it to a sequantally ordered `output_XXXX` folder), but the drawback of this is that moving the data has to occur after the fact, which adds costs and makes it harder to implement in a way that is safe with respect to crashes.

charleskawczynski:
> The use of junction points is internal in julia (they are still created by the `symlink` function). Junction points are symlinks with two extra restrictions: they require full paths and only work with directories. Users of the `OutputPathGenerator` do not have to specify full paths: the PR prepends them automatically when they are needed.

Ah, ok. Sounds good.

> For the point of view of users, the only difference between OSs is that if your OS doesn't support actual symlinks, you won't be able to move the `base_output_directory` (without breaking the link).

👍🏻

> This is the idea I came up with trying to find a solution that
> 
> * doed not touch existing data
> * provides a fixed and predicatable output path for the newest data
> * has nearly zero performance cost
> 
> The way I came up with to do this without symlinks involved moving data (e.g., the output is always is a new temporarly folder, but then we "file" it to a sequantally ordered `output_XXXX` folder), but the drawback of this is that moving the data has to occur after the fact, which adds costs and makes it harder to implement in a way that is safe with respect to crashes.

Ok, that sounds reasonable. Can we highlight these design goals in the documentation?

src/OutputPathGenerator.jl
```diff
@@ -209,8 +219,23 @@ function generate_output_path(::ActiveLinkStyle, output_path; context = nothing)
...
-        # In Windows, symlinks must be explicitly declared as referring to a directory or not
-        symlink("output_$next_counter_str", active_link, dir_target = true)
+        # On Windows, creating symlinks might require admin privileges. This depends on the
+        # version of Windows and some of its settings (e.g., if "Developer Mode" is enabled).
+        # So, we first try creating a symlink. If this fails, we resort to creating a NTFS
+        # junction. This is almost the same as a symlink, except it requires absolute paths.
+        # In general, relative paths would be preferable because they make the output
+        # completely relocatable (whereas absolute paths are not).
+        try
+            symlink("output_$next_counter_str", active_link, dir_target = true)
```
charleskawczynski, line 229: When/how might this fail?

I'm generally in favor of avoiding try-catch if possible.

src/OutputPathGenerator.jl
```diff
@@ -209,8 +219,23 @@ function generate_output_path(::ActiveLinkStyle, output_path; context = nothing)
...
+        catch e
+            if e isa Base.IOError && Base.uverrorname(e.code) == "EPERM"
+                active_link = abspath(active_link)
+                dest_active_link = abspath("output_$next_counter_str")
+                symlink(dest_active_link, active_link, dir_target = true)
+            else
+                rethrow(e)
+            end
```
charleskawczynski, line 237: Can we test that this `catch` block is exercised in CI?

src/OutputPathGenerator.jl
Sbozzolo: I agree, I also dislike try-catch but I don't think that this is possible without it. This will fail on specific versions of windows, but the failure comes from the output of a system call, so I don't think it can be done without trying.

src/OutputPathGenerator.jl
Sbozzolo: I would like to, but I don't know how. The runner in CI has a version of windows where we can create symlinks and this is catching specifically the case where symlinks cannot be created because of permission errors.

Sbozzolo:
> Ok, that sounds reasonable. Can we highlight these design goals in the documentation?

Yes, added!

src/OutputPathGenerator.jl
charleskawczynski: Ok, I guess we can proceed with trial-error. Let's try this out and we can open issues if something isn't still working.

charleskawczynski (approved):

PR #60, opened 2024-05-21
Add downgrade test

Sbozzolo:
To check for lower bound compats


.github/workflows/Downgrade.yml
```diff
@@ -0,0 +1,34 @@
...
+        version: ['1']
+    steps:
+      - uses: actions/checkout@v4
+      - uses: julia-actions/setup-julia@latest
+        with:
+          version: ${{ matrix.version }}
+      - uses: julia-actions/cache@v2
+      - uses: julia-actions/julia-downgrade-compat@v1
+        with:
+          skip: Artifacts, Dates, Profile, Test
```
juliasloan25, line 32: I don't see Profile in the top-level or test Projects, maybe we can remove it here?

juliasloan25 (approved):
Thanks Gabriele, this looks super helpful! I think it would be helpful to have this in ClimaLand and ClimaCoupler too, so I may add it there

.github/workflows/Downgrade.yml
Sbozzolo: You are right. Thank you!

Sbozzolo:
> Thanks Gabriele, this looks super helpful! I think it would be helpful to have this in ClimaLand and ClimaCoupler too, so I may add it there

FYI, it may be painful to set up because you have to find the correct lower bounds for the Project.toml (and the only way I know how to is with trial and error)

PR #63, opened 2024-05-27
Change ActiveLinkStyle to return the folder instead of link

Sbozzolo:
Soft links are not like hard links or folders. Sometimes, they are treated differently by the operating system (and Julia). As a result, for some operations, an extra step is required to work with the output produced by the ActiveStyle in OutputPathGenerator.

This PR changes the returned path to be the newly created folder instead of the link. The link is still being created and managed, which means that users can still assume that their output can always be found in `output_active`, but the path that is presented to Julia code that uses `OutputPathGenerator` is now the path of the newly created folder. This has the advantage that the path is a normal folder, possibly simplifying certain operations.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,13 @@
 ClimaUtilities.jl Release Notes
 ===============================
 
+v0.1.8
+------
+
+- `generate_output_path(..., ::ActiveLinkStyle)` now returns the folder instead
+  of the link. Links are still being created and managed. PR
+  [#63](https://github.com/CliMA/ClimaUtilities.jl/pull/63)
+
 v0.1.7
 ------
 
@@ -24,3 +31,5 @@ v0.1.3
 -------
 - Add `DataStructures` module containing `LRUCache` object. PR [#35](https://github.com/CliMA/ClimaUtilities.jl/pull/35)
 - Add `OutputPathGenerator`. PR [#28](https://github.com/CliMA/ClimaLand.jl/pull/28)
+
+[badge-💥breaking]: https://img.shields.io/badge/💥BREAKING-red.svg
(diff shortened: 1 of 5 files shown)


NEWS.md
```diff
@@ -1,6 +1,11 @@
 ClimaUtilities.jl Release Notes
 ===============================
 
+v0.1.8
+------
+
+- `OutputPathGenerator` `ActiveLinkStyle` now returns the folder instead of the link. Links are still being created and managed. PR [#63](https://github.com/CliMA/ClimaUtilities.jl/pull/63)
```
charleskawczynski, line 7:
```suggestion
- `generate_output_path(..., ::ActiveLinkStyle)` now returns the folder instead of the link. Links are still being created and managed. PR [#63](https://github.com/CliMA/ClimaUtilities.jl/pull/63)
```

docs/src/outputpathgenerator.md
```diff
@@ -42,12 +42,13 @@ results.
...
-* If `data` doesn't exist, the module creates it and returns
-  `data/output_active`. This link points to the newly created subfolder
-  `data/output_0000`.
+* If `data` doesn't exist, the module creates it and returns `data/output_0000`.
+  In doing this, a link `data/output_active` to `data/output_0000` is created so
+  that you can always access your data in `data/output_active`.
 * If `data` exists and contains an `output_active` link pointing to
   `data/output_0005`, the module creates a new subfolder `data/output_0006` and
-  updates `output_active` to point to it.
+  updates `output_active` to point to it. `generate_output_path` returns
```
charleskawczynski, line 50: Can we change this to:

```
* If `data` exists and contains an `output_active` link pointing to
  `data/output_0005`, `output_active` is updated to point to a new subfolder
called `data/output_0006`.
```

charleskawczynski (approved):
Looks good, I added a few suggestions

Sbozzolo:
> Looks good, let's try this out, I added a few suggestions

Thank you!

Issue #65, opened 2024-06-04
TempestRegridder can return Spaces that are different from the input space


Sbozzolo:
`TempestRegridder` is not GPU/MPI compatible. For GPU/MPI runs, TempestRegridder creates a new Singleton CPU Space where operations are performed, and then distributes back the space. There is no guarantee/check that resulting space is the same as the input one.

This resulted in #53 

### Potential solution
- [x] delete `swap_space`
  - this function bypasses ClimaCore and rearranges internals in a potentially inconsistent way that we want to avoid
- [x] add a check in `read_from_hdf5` that `axes(field) == space`
  - I think this will be false with the code as we have it now - need the below to make sure this is true
- [x] construct undistributed space with or without `spacefillingcurve` depending on input space
  - want these spaces to be identical except for the context
- [ ] `hdwrite_regrid...` should call `write_to_hdf5` using a field on the original space, rather than the undistributed space
  - can call CCTR on space constructed with spacefillingcurve
  - should be okay as long as the undistributed space we construct only differs from the original space in the context (i.e. topology.elemorder is the same, either both using or both not using `spacefillingcurve`)

PR #66, opened 2024-06-05
conditionally use spacefillingcurve

juliasloan25:
## Purpose 
closes https://github.com/CliMA/ClimaUtilities.jl/issues/65

related https://github.com/CliMA/ClimaCoupler.jl/issues/764

Tested in ClimaCoupler in singleton CPU, MPI, and GPU cases: https://buildkite.com/clima/climacoupler-ci/builds/4019#0190282c-6ea6-4f74-a964-70ea9749a3b4

Tested in ClimaLand: https://buildkite.com/clima/climaland-ci/builds/4314

## Content

----

### DIFF
--- a/docs/src/regridders.md
+++ b/docs/src/regridders.md
@@ -34,6 +34,9 @@ computational domains. `TempestRegridder` performs all the interpolation ahead
 of time and saves the regridded fields to HDF5 files that can be read during the
 simulation.
 
+Currently, `TempestRegridder` does not support regridding on 3D domains.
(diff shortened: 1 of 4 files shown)


Sbozzolo:
I think we should use the same mesh as the input space (namely, with or without spacefilling curve) and not hardcode either.

Sbozzolo:
Could you also add a check to ensure that the final Space we return is the same as the input one? (Closing #65)

Sbozzolo (approved):

PR #69, opened 2024-06-17
Reorder extensions

### LINKED ISSUE #68: Avoid circular dependencies in extensions
https://github.com/CliMA/ClimaUtilities.jl/blob/18bad9cd95d84f84ceed7bcccfb6c4d5886a8686/Project.toml#L19-L28

Extensions are defined to mirror the modules defined in `src`. However, this leads to circular dependencies, where some extensions trigger loading of others. This is not properly handled by the precompiling logic in Julia as of Julia 1.9-1.11 (fixed in master, upcoming to Julia 1.12, or maybe it will be backported at some point).

Sbozzolo:
Extensions are defined to mirror the modules defined in src. However, this leads to circular dependencies, where some extensions trigger the loading of others. This is not properly handled by the precompiling logic in Julia as of Julia 1.9-1.11 (fixed in master, upcoming to Julia 1.12, or maybe it will be backported at some point).

Here, I change the way extensions are organized to match the packages that are needed to load them. In this, I still maintain the abstraction that each module in `src` has its implementation in `ext/XYZExt.jl`.

I also added the prefix `ClimaUtilities` to everything to make sure that it won't clash with anything.

Closes #68

### DIFF
--- a/ext/ClimaUtilitiesClimaCommsExt.jl
+++ b/ext/ClimaUtilitiesClimaCommsExt.jl
@@ -1,0 +1,5 @@
+module ClimaUtilitiesClimaCommsExt
+
+include("MPIUtilsExt.jl")
+
+end
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,12 @@
 ClimaUtilities.jl Release Notes
 ===============================
 
+v0.1.9
+------
+
+- Extensions are internally reorganized, removing precompilation errors. PR
+  [#69](https://github.com/CliMA/ClimaUtilities.jl/pull/69)
+
 v0.1.8
 ------
 
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "ClimaUtilities"
 uuid = "b3f4f4ca-9299-4f7f-bd9b-81e1242a7513"
 authors = ["Gabriele Bozzola <gbozzola@caltech.edu>", "Julia Sloan <jsloan@caltech.edu>"]
-version = "0.1.8"
+version = "0.1.9"
 
 [deps]
 Artifacts = "56f22d72-fd6d-98f1-02f0-08ddc0907c33"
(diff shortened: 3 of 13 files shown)


ext/ClimaUtilitiesClimaCommsExt.jl
```diff
@@ -0,0 +1,5 @@
+module ClimaUtilitiesClimaCommsExt
+
+include("MPIUtilsExt.jl")
```
charleskawczynski, line 3: Is it intended for this to be a module within a module?

charleskawczynski (approved):
I'm overall in favor of reorganizing the extensions if it fixes the error.

One comment I have is: can we make sure that we use `TitleCase` for module files and `snake_case` for non-module files?

Also, I was kind of wondering, what is the utility of splitting up two extensions that are triggered by the same set of packages being loaded? I think it might make more sense to join them together.

Sbozzolo:
> I'm overall in favor of reorganizing the extensions if it fixes the error.
> 
> One comment I have is: can we make sure that we use `TitleCase` for module files and `snake_case` for non-module files?

> Also, I was kind of wondering, what is the utility of splitting up two extensions that are triggered by the same set of packages being loaded? I think it might make more sense to join them together.

I tried explaining the reasoning in my addendum to the readme. (They are all module files, so the convention on names is properly applied here)

Let me expand here:

The way ClimaUtilities uses extensions is slightly different from other packages. ClimaUtilities has several modules and the entirety of the implementation in extensions. The relevant abstraction is that there's a module in `src` and its implementation is in a module in `ext`. This essentially follows what we would do in an ideal world where we could do partial loading, where everything would be in its own module in `src`. Different modules are logically distinct and should be kept in different namespaces, only interacting with each other through the public keywords defined in `src`. The fact that some modules may be loaded by the same set of packages is accidental and should not be relied upon, instead, each module should be allowed to be developed independently (with a possibly changing set of dependencies/trigger packages).

charleskawczynski (approved):

PR #72, opened 2024-07-02
Preallocate regridded spaces

Sbozzolo:
Reducing allocations. And expose `regridded_snapshot!`

### DIFF
--- a/ext/TimeVaryingInputsExt.jl
+++ b/ext/TimeVaryingInputsExt.jl
@@ -15,7 +15,7 @@ import ClimaUtilities.TimeVaryingInputs: NearestNeighbor, LinearInterpolation
 
 import ClimaUtilities.DataHandling
 import ClimaUtilities.DataHandling:
-    DataHandler, previous_time, next_time, regridded_snapshot, available_times
+    DataHandler, previous_time, next_time, regridded_snapshot!, available_times
 
 # Ideally, we should be able to split off the analytic part in a different
 # extension, but precompilation stops working when we do so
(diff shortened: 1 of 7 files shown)


ext/TimeVaryingInputsExt.jl
```diff
@@ -75,6 +76,9 @@ struct InterpolatingTimeVaryingInput23D{
     """Range of times over which the interpolator is defined. range is always defined on the
     CPU. Used by the in() function."""
     range::R
+
```
kmdeck, line 79: extra newline?

ext/TimeVaryingInputsExt.jl
```diff
@@ -75,6 +76,9 @@ struct InterpolatingTimeVaryingInput23D{
...
+    """Preallocated spaces where to do regridding"""
```
kmdeck, line 80: "Preallocated memory for storing regridded fields" (only because "space" also has another meaning in a lot of what we do)

ext/TimeVaryingInputsExt.jl
```diff
@@ -179,9 +190,13 @@ function TimeVaryingInputs.evaluate!(
     t0, t1 =
         previous_time(itp.data_handler, time), next_time(itp.data_handler, time)
     coeff = (time - t0) / (t1 - t0)
-    dest .=
-        (1 - coeff) .* regridded_snapshot(itp.data_handler, t0) .+
-        coeff .* regridded_snapshot(itp.data_handler, t1)
+
+    field_prev = itp.preallocated_regridded_fields[1]
```
kmdeck, line 194: extremely minor, but could align naming (field0 with t0, or field_prev with t_prev). as you prefer!

kmdeck (approved):
nice change! thanks!

ext/TimeVaryingInputsExt.jl
Sbozzolo: The fields in this struct are already spaced with one extra line to make reading the docstring a little easier

PR #75, opened 2024-07-11
Preallocate GPU interpolant

Sbozzolo:
`Adapt` allocates new arrays on the GPU. With this PR, I change `Adapt` to be in place.


ext/InterpolationsRegridderExt.jl
```diff
@@ -69,7 +75,12 @@ function Regridders.InterpolationsRegridder(
         end
     end
 
-    return InterpolationsRegridder(target_space, coordinates, extrapolation_bc)
+    return InterpolationsRegridder(
+        target_space,
+        coordinates,
+        extrapolation_bc,
+        zeros(target_space),
```
charleskawczynski, line 82: Why not just use adapt here?

PR #76, opened 2024-07-23
Run CI on 1.11

Sbozzolo:
Update CI to run on 1.11 too


juliasloan25:
CI is failing with v1.11 - maybe we should get it running before we merge this

Sbozzolo:
> CI is failing with v1.11 - maybe we should get it running before we merge this

Yes, I didn't realize this, but it makes sense. Our lower bounds of some packages are not compatible with julia 1.11 (e.g., CUDA = 5 is not, we need CUDA 5.5). I am not entirely sure how to solve this if not bumping the compats.

juliasloan25 (approved):

PR #77, opened 2024-07-23
Add support for boundary conditions in TimeVaryingInputs

Sbozzolo:
This commit implements boundary conditions for extrapolation of TimeVaryingInputs.

Boundary conditions are specified with subtypes of `AbstractInterpolationMethod` and passed to the `InterpolationMethod`.

Currently, three boundary conditions are implemented:

- `ThrowError`. This recovers the previous behavior and essentially means that extrapolation in time is not allowed
- `Clamp`. This clamps extrapolated values to the closest edge.
- `PeriodicCalendar`. More on this below.

`ThrowError` and `Clamp` are implemented at the level of the main `evaluate!`, while `PeriodicCalendar` requires more infrastructure.

`PeriodicCalendar` lays the groundwork for allowing a `TimeVaryingInput` to repeat itself in a way that is somewhat aware of the time duration.

Consider the example where we have data on the 15th of every month for the year 2024: 01/15/24, 02/15/24, ....., 12/15/24

`PeriodicCalendar` "repeats" this year as it the data we had was 12/15/23, 01/15/24, 02/15/24, ....., 12/15/24, 01/15/25, ...
where every year is identical to 2024.

This is not fully implemented yet, reason being that months have different number of days, which is an extra complication. Instead, this commit requires the data to be uniformly sampled. This is needed because we need to know what comes after t_end.

I will add support for dates next.

### DIFF
--- a/docs/src/inputs.md
+++ b/docs/src/inputs.md
@@ -75,6 +75,28 @@ albedo_tv = TimeVaryingInputs.TimeVaryingInput("cesem_albedo.nc", "alb", target_
 # model developers do not have to worry about anything :)
 ```
 
+`TimeVaryingInput`s can have multiple boundary conditions for extrapolation. By
+default, the `Throw` condition is used, meaning that interpolating onto a
+point that is outside the range of definition of the data is not allowed. Other
+boundary conditions are allowed. For instance, the `PeriodicCalendar` condition
+"repeats" the data over and over. With this boundary condition, if the data is
+defined over a period of time from `t0` to `t1` (e.g., 1 and 5), interpolating
+over `t > t1` (e.g., 7) is equivalent to interpolating to `t*` where `t*` is the
+modulus of `t` and the range (3 in this case). This can be used to repeat one
+year of data (suppose you have data defined on the 15th of every month for the a
+given year Y. With `PeriodicCalendar`, you evaluate the data for any year using
+data for Y). `PeriodicCalendar` requires the data to be uniformly spaced in
+time. To enable this boundary condition, pass
+`LinearInterpolation(PeriodicCalendar())` to the `TimeVaryingInput` (or
+`NearestNeighbor(PeriodicCalendar())`).
+
+> Note: this `PeriodicCalendar` is different from what you might be used to, where the
+> identification is `t1 = t0`. Here, we identify `t1 + dt = t0`. This is so that
+> we can use it to repeat calendar data.
+
+Another option is `Flat`: when interpolating outside of the range of
+definition, return the value of the closest boundary.
+
 ## `SpaceVaryingInputs`
 
 > This extension is loaded when loading `ClimaCore` is loaded. In addition to
@@ -124,9 +146,14 @@ albedo_from_file = SpaceVaryingInputs.SpaceVaryingInput("cesm_albedo.nc", "alb",
 ```@docs
 ClimaUtilities.SpaceVaryingInputs.SpaceVaryingInput
 ClimaUtilities.TimeVaryingInputs.AbstractInterpolationMethod
+ClimaUtilities.TimeVaryingInputs.AbstractInterpolationBoundaryMethod
 ClimaUtilities.TimeVaryingInputs.NearestNeighbor
 ClimaUtilities.TimeVaryingInputs.LinearInterpolation
+ClimaUtilities.TimeVaryingInputs.Throw
+ClimaUtilities.TimeVaryingInputs.PeriodicCalendar
+ClimaUtilities.TimeVaryingInputs.Flat
 ClimaUtilities.TimeVaryingInputs.evaulate!
+ClimaUtilities.TimeVaryingInputs.extrapolation_bc
 Base.in
 Base.close
 ```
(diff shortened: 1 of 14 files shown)


src/TimeVaryingInputs.jl
```diff
@@ -50,9 +50,54 @@ Defines how to perform interpolation.
 
 Not all the TimeVaryingInputs support all the interpolation methods (e.g., no interpolation
 methods are supported when the given function is analytic).
+
+`AbstractInterpolationMethod`s have to implement a `boundary_conditions` field.
 """
 abstract type AbstractInterpolationMethod end
 
+"""
```
kmdeck, line 58: For space varying inputs, we have "Extrapolation BC", with similar options, but different naming...
ThrowError vs Interpolations.Throw
Clamp vs Interpolations.Flat
PeriodicCalendar vs Interpolations.Periodic
e.g.:https://github.com/CliMA/ClimaUtilities.jl/blob/b8f6840c92eb0e6bca18ca6de3bbdc95a2826f5a/ext/InterpolationsRegridderExt.jl#L50

we also seem to be calling it extrapolation and interpolation interchangeably.

If possible, I think we should unify the naming of the bc types for space and time (if they are the same concept) and use extrapolation instead of interpolation

kmdeck: Oh - I see that PeriodicCalendar is not the same as Periodic!

src/DataHandling.jl
```diff
@@ -55,4 +55,6 @@ function regridded_snapshot end
 
 function regridded_snapshot! end
 
+function dt end
```
kmdeck, line 58: since `dt` often refers to the timestep value itself, should this be `get_dt`?

kmdeck: also, maybe `get_data_dt`? to avoid confusion with simulation dt?

ext/DataHandlingExt.jl
```diff
@@ -263,6 +282,7 @@ function DataHandling.previous_time(
     else
         index = searchsortedfirst(data_handler.available_dates, date) - 1
     end
+    index < 1 && error("date/time not available in the data_handler")
```
kmdeck, line 285: Maybe the error message could be "Date/time prior to any date/time found in data handler".

is this only used with `ThrowError` option?

ext/DataHandlingExt.jl
```diff
@@ -285,6 +307,8 @@ function DataHandling.next_time(data_handler::DataHandler, date::Dates.DateTime)
     else
         index = searchsortedfirst(data_handler.available_dates, date)
     end
+    index > length(data_handler.available_dates) &&
+        error("date/time not available in the data_handler")
```
kmdeck, line 311: Maybe the error message could be "Date/time after any date/time found in data handler"

ext/TimeVaryingInputs0DExt.jl
```diff
@@ -5,10 +5,13 @@ import ClimaCore: ClimaComms
...
 
-import ClimaUtilities.Utils: searchsortednearest, linear_interpolation
+import ClimaUtilities.Utils:
+    searchsortednearest, linear_interpolation, wrap_time, isequispaced
 import ClimaUtilities.TimeVaryingInputs:
     AbstractInterpolationMethod, AbstractTimeVaryingInput
-import ClimaUtilities.TimeVaryingInputs: NearestNeighbor, LinearInterpolation
+import ClimaUtilities.TimeVaryingInputs:
+    NearestNeighbor, LinearInterpolation, ThrowError, Clamp, PeriodicCalendar
+import ClimaUtilities.TimeVaryingInputs: boundary_conditions
```
kmdeck, line 14: yes, I think here we called in `extrapolation_bc` in for spatial regridding..

docs/src/inputs.md
```diff
@@ -75,6 +75,27 @@ albedo_tv = TimeVaryingInputs.TimeVaryingInput("cesem_albedo.nc", "alb", target_
...
+boundary conditions are allowed. For instance, the `PeriodicCalendar` condition
+"repeats" the data over and over. With this boundary condition, if the data is
+defined over a period of time from `t0` to `t1` (e.g., 1 and 5), interpolating
+over `t > t1` (e.g., 7) is equivalent to interpolating to `t*` where `t*` is the
+modulus of `t` and the range (3 in this case). This can be used to repeat one
+year of data (suppose you have data defined on the 15th of every month for the a
+given year Y. With `PeriodicCalendar`, you evaluate the data for any year using
+data for Y). `PeriodicCalendar` requires the data to be uniformly spaced in
+time. To enable this boundary condition, pass
+`LinearInterpolation(PeriodicCalendar())` to the `TimeVaryingInput`.
```
kmdeck, line 90: or NearestNeighbor(PeriodicCalendar())?

kmdeck: double checking if we are restricted to just LinearInterpolation for any of these boundary conditions

Sbozzolo: They are all supported

Sbozzolo:
Superseded by #84

PR #84, opened 2024-08-01
Add support for full PeriodicCalendar

### LINKED ISSUE #73: Add support for boundary conditions in time interpolation
As of right now, extrapolation for `TimeVaryingInput`s is not possible. There are different reasonable extrapolation behaviors different users might want:

- [x] `ThrowError`, extrapolation is not allowed
- [x] `Clamp`, the edge boundary value is used
- [x] `PeriodicCalendar`, we wrap data around

Sbozzolo:
This PR implements boundary conditions for extrapolation of TimeVaryingInputs.

Boundary conditions are specified with subtypes of `AbstractInterpolationMethod` and passed to the `InterpolationMethod`.

Currently, three boundary conditions are implemented:

- `ThrowError`. This recovers the previous behavior and essentially means that extrapolation in time is not allowed
- `Clamp`. This clamps extrapolated values to the closest edge.
- `PeriodicCalendar`. More on this later.

`ThrowError` and `Clamp` are implemented at the level of the main `evaluate!`, while `PeriodicCalendar` requires more infrastructure.

`PeriodicCalendar` allows `TimeVaryingInput`s to repeat themself in a way that is somewhat aware of the time duration.

Consider the example where we have data on the 15th of every month for the year 2024: 01/15/24, 02/15/24, ....., 12/15/24

`PeriodicCalendar` "repeats" this year as it the data we had was 12/15/23, 01/15/24, 02/15/24, ....., 12/15/24, 01/15/25, ...
where every year is identical to 2024.

`PeriodicCalendar` can take two arguments: `period`, and `repeat_date`. `period` is a unitary `Dates.DatePeriod` (e.g.,
`Dates.Year(1)`) that defines what the periodicity should be. `repeat_date` defines the date that should be repeated with the given `period` (e.g., `repeat_date = Dates.Date(1993, 11, 19)` and period `Dates.Year(1)` repeats the year 1993).

When arguments are not provided, `PeriodicCalendar` assumes that the data is defined on one period.

This PR also includes two house keeping commits.

Closes #73


NEWS.md
```diff
@@ -1,17 +1,25 @@
...
+  [#84](https://github.com/CliMA/ClimaUtilities.jl/pull/84)
+- Increased alloctions in regridding. `read!` method removed. PR
+  [#84](https://github.com/CliMA/ClimaUtilities.jl/pull/84)
+
 v0.1.11
 ------
 
-- Reduced alloctions in regridding. New method `regridded_snapshot!`. PR
-  [#72](https://github.com/CliMA/ClimaUtilities.jl/pull/72)
+- Reduced alloctions in regridding. New method `read!`. PR
```
kmdeck, line 15: allocations (here and below)

docs/src/inputs.md
```diff
@@ -124,9 +155,14 @@ albedo_from_file = SpaceVaryingInputs.SpaceVaryingInput("cesm_albedo.nc", "alb",
 ```@docs
 ClimaUtilities.SpaceVaryingInputs.SpaceVaryingInput
 ClimaUtilities.TimeVaryingInputs.AbstractInterpolationMethod
+ClimaUtilities.TimeVaryingInputs.AbstractInterpolationBoundaryMethod
 ClimaUtilities.TimeVaryingInputs.NearestNeighbor
 ClimaUtilities.TimeVaryingInputs.LinearInterpolation
+ClimaUtilities.TimeVaryingInputs.Throw
+ClimaUtilities.TimeVaryingInputs.PeriodicCalendar
+ClimaUtilities.TimeVaryingInputs.Flat
 ClimaUtilities.TimeVaryingInputs.evaulate!
```
kmdeck, line 164: evaluate!

ext/DataHandlingExt.jl
```diff
@@ -234,6 +225,21 @@ function DataHandling.available_dates(data_handler::DataHandler)
     return data_handler.available_dates
 end
 
+"""
+    dt(data_handler::DataHandler)
+
+Return the time interval between for the data in `data_handler`.
```
kmdeck, line 231: "Return the time interval between data points in `data_handler`"?

ext/DataHandlingExt.jl
```diff
@@ -243,24 +249,47 @@ Convert the given time to a calendar date.
...
+    data_handler::DataHandler,
+    time::AbstractFloat,
+)
     return data_handler.reference_date +
            Second(data_handler.t_start) +
            Second(time)
 end
 
+"""
+    date_to_time(data_handler::DataHandler, time::AbstractFloat)
```
kmdeck, line 262: I think this is fine - I assume from context that `time` refers to simulation time?

ext/TimeVaryingInputs0DExt.jl
```diff
@@ -108,13 +114,26 @@ end
...
 
+    if extrapolation_bc(method) isa PeriodicCalendar
+        if extrapolation_bc(method) isa PeriodicCalendar{Nothing}
+            isequispaced(times) || error(
+                "PeriodicCalendar() boundary condition cannot be used because data is defined at non uniform intervals of time",
+            )
+        else
+            # We have the period in PeriodicCalendar
+            error(
+                "PeriodicCalendar(period) is not supported when the input data is 1D",
```
kmdeck, line 132: Why is this?

and, I think in this context the TimeVaryingInput0D refers to spatial dimensions (i.e. a time series at a point). Should the 1D here be 0D?

ext/TimeVaryingInputs0DExt.jl
```diff
@@ -161,14 +198,30 @@ function TimeVaryingInputs.evaluate!(
...
+        if time >= t_end
+            dest .= itp.vals[end]
+        else
+            time <= t_init
+            dest .= itp.vals[begin]
+        end
+        return nothing
+    elseif extrapolation_bc(itp.method) isa PeriodicCalendar
+        t_init, t_end = itp.range
+        dt = itp.times[begin + 1] - itp.times[begin]
```
kmdeck, line 212: Could you use `DataHandling.dt` here?

ext/TimeVaryingInputs0DExt.jl
```diff
@@ -137,6 +156,24 @@ function TimeVaryingInputs.evaluate!(
...
+        if time >= t_end
+            dest .= itp.vals[end]
+        else
+            time <= t_init
+            dest .= itp.vals[begin]
+        end
+        return nothing
+    elseif extrapolation_bc(itp.method) isa PeriodicCalendar{Nothing}
+        t_init, t_end = itp.range
+        dt = itp.times[begin + 1] - itp.times[begin]
```
kmdeck, line 170: here too

ext/TimeVaryingInputs0DExt.jl
```diff
@@ -137,6 +156,24 @@ function TimeVaryingInputs.evaluate!(
 )
     # Nearest neighbor interpolation: just pick the values corresponding to the entry in
     # itp.times that is closest to the given time.
+    if extrapolation_bc(itp.method) isa Flat
```
kmdeck, line 159: since this code is repeated below, is it worth making a helper function for it?

kmdeck: maybe it is too different with the linear vs nn interpolations to do this conveniently

ext/TimeVaryingInputs0DExt.jl
Sbozzolo: This is for the TimeVaryingInputs0D, so there is no DataHandling here.

ext/DataHandlingExt.jl
Sbozzolo: I mean, at the end of the day, yes. But more fundamentally, `time` is the offset in seconds between `reference_date` and `date` given `t_start`.

ext/TimeVaryingInputs0DExt.jl
Sbozzolo: 1D because the input data is a 1D array `times` vs `values`.

We do not support `PeriodicCalendar(period)` (note: with a passed `period`, we support it without) because the input `times` array has no notion of dates.

ext/TimeVaryingInputs0DExt.jl
Sbozzolo: No, because this is `TimeVaryingInput0D`, so there is no `DataHandling`.

ext/TimeVaryingInputs0DExt.jl
Sbozzolo: I can do it if you want :)

src/Utils.jl
```diff
@@ -43,4 +48,271 @@ function linear_interpolation(indep_vars, dep_vars, indep_value)
...
+0.2000000000000001
+```
+"""
+function wrap_time(time, t_init, t_end; extend_past_t_end = false, dt = nothing)
+    extend_past_t_end &&
+        isnothing(dt) &&
+        error("dt required with extend_past_t_end = true")
+    if extend_past_t_end
+        t_end = t_end + dt
+        return wrap_time(time, t_init, t_end; extend_past_t_end = false)
```
kmdeck, line 175: whoa! I didnt know you could call a function within itself.

That said, I had the question of why we have the two modes when they can be written as a single function with t_end = t_end+dt in one case vs just t_end in the other.

i.e., the function could be wrap_time(t, t_init, t_end) where t_end = t_init given the periodicity, even if t_end is not the time of the last data point.

kmdeck: or it could even be `wrap_time(t, t_init, period)`

src/Utils.jl
```diff
@@ -43,4 +48,271 @@ function linear_interpolation(indep_vars, dep_vars, indep_value)
...
+1993-01-01T00:00:00
+
+julia> beginningofperiod(Date(1993, 11, 19), Month(1))
+1993-11-01T00:00:00
+
+julia> beginningofperiod(Date(1993, 11, 19), Week(1))
+1993-11-15T00:00:00
+
+julia> beginningofperiod(Date(1993, 11, 19), Day(1))
+1993-11-15T00:00:00
```
kmdeck, line 203: Maybe this should be `1993-11-19T00:00:00`, and not `15`?

src/Utils.jl
```diff
@@ -43,4 +48,271 @@ function linear_interpolation(indep_vars, dep_vars, indep_value)
...
+        error("Unsupported period type")
+    end
+end
+
+"""
+    bounding_dates(dates, target_date::Union{Date, DateTime}, period::DatePeriod)
+
+Identify bounding dates within `dates` that share the same `period` as `target_date`.
+
+An example will make clear what this function really does:
```
kmdeck, line 276: the example is great, but perhaps it could be helpful to add a use case? 

"This function is used, for example, when data is available in a single file for many years, but a period of only one year is wished to be simulated repeatedly. In this case, dates would be the dates of the data over all the years, the target date would be any date within the year wished to be simulated, and the period would be one year."?

and often the target_date would be the reference date?

ext/TimeVaryingInputs0DExt.jl
kmdeck: just to be clear, does this mean that when we run site-level runs the time always is in seconds measured relative to t_start? I think in all of our examples we read in the data and pass in a time array like this.

ext/TimeVaryingInputs0DExt.jl
kmdeck: eh, up to you.

ext/TimeVaryingInputs0DExt.jl
kmdeck: and this is because we pass in the arrays having read in the data already (no issues storing all the data), whereas with the global runs we pass the file path and read in chunks, right?

ext/TimeVaryingInputsExt.jl
```diff
@@ -147,11 +167,110 @@ function TimeVaryingInputs.evaluate!(
...
+            time <= t_init
+            regridded_snapshot!(dest, itp.data_handler, t_init)
+        end
+    else
+        TimeVaryingInputs.evaluate!(dest, itp, time, itp.method)
+    end
     return nothing
 end
 
+function _time_range_dt_dt_e(itp::InterpolatingTimeVaryingInput23D)
```
kmdeck, line 187: what does this function name mean?

ext/TimeVaryingInputsExt.jl
```diff
@@ -147,11 +167,110 @@ function TimeVaryingInputs.evaluate!(
...
+Return time, t_init, t_end, dt, dt_e.
+
+Implementation details
+======================
+
+Okay, how are we implementing PeriodicCalendar?
+
+There are two modes, one with provided period and repeat_date, and the other without. We
+reduce the second case to the first one. So, let's start by looking at the second case.
+Then, we will look how we reduce the first case to the second one.
```
kmdeck, line 238: in the sentence just prior, you say "reduce second case to first one".

ext/TimeVaryingInputsExt.jl
kmdeck: nevermind!

ext/TimeVaryingInputs0DExt.jl
Sbozzolo: Yes, right

src/TimeVaryingInputs.jl
```diff
@@ -50,9 +52,89 @@ Defines how to perform interpolation.
...
+data).
+
+> Note: Passing a period is not supported by all the interpolators (e.g., when reading from
+  1D files).
+
+Second, if no period is provided, when interpolating outside of range, restart from the beginning.
+
+For example, if the data is defined from t0 = 0 to t1 = 10, extrapolating at t=13 is
+equivalent to interpolating at t=2. In practice, we identify `t1 + dt` to be `t0` again.
+This is different from what you might be used to for `Periodic`, where the identification is
```
kmdeck, line 98: reference to `Periodic` - maybe should read something like `This is different from what you might be used to for periodic boundary conditions, where the identification is t1 = t0`

ext/TimeVaryingInputs0DExt.jl
Sbozzolo: Added some helper functions

kmdeck (approved):
Youve thought about this a lot more than me, and it may just be the nature of it, but parts of this are quite complex.

- if the two cases reduce to each other, can we change the interface so that only one option is available to the user? It seems like in both cases the user could supply a period and start date/time. I.e., the period could be t_end+dt - t_init, and the repeat date (which currently can be anything) could be t_init?
- do we require two options because we need to support  the TVI 0D, which only works with times and not dates? Or because we really want to support strictly periodic BC (t_end = t_init)?

Sbozzolo:
> Youve thought about this a lot more than me, and it may just be the nature of it, but parts of this are quite complex.
> 
> * if the two cases reduce to each other, can we change the interface so that only one option is available to the user? It seems like in both cases the user could supply a period and start date/time. I.e., the period could be t_end+dt - t_init, and the repeat date (which currently can be anything) could be t_init?
> * do we require two options because we need to support  the TVI 0D, which only works with times and not dates? Or because we really want to support strictly periodic BC (t_end = t_init)?

The two user-facing options are:
1. You want to repeat the entire dataset over and over
2. Or, you want to provide a specific date and period and repeat only data (only for TimeVaryingInputs that read from NetCDF files)
2 reduces to 1 in implementation but not in what it actually does, so maybe I should clarify that.

I think having both options is reasonable and the only difference is in the constructor of `PeriodicCalendar`. For the first option, it is just `PeriodicCalendar()`, for the second you provide the date and period.

src/Utils.jl
Sbozzolo: You are right! Indeed, we don't really need the second mode.

PR #85, opened 2024-08-05
Add LinearPeriodFillingInterpolation

### LINKED ISSUE #74: Add support for time interpolation with jump
For aerosol data (and probably other cases too), regular time interpolation is not the best. More specifically, we don't want to interpolate across months, instead, we interpolate over the same months but across years.

For example, if we have monthly data for various years, we want to interpolate on Jan 2015 using data from Jan 2014 and Jan 2016 (instead of Dec 2014 and Feb 2015) (for linear interpolation).

Sbozzolo:
This PR adds support for period-filling interpolation. From the docstring:

> Perform a period-filling linear interpolation between the two neighboring points.
> 
> This interpolation is period-filling because it ensures that there is no gap on input data
> larger than the given period.
> 
> This interpolation method is best explained with an example. Let us assume data is defined
> on 03/01/1985, 05/01/1985, 11/02/1985, ... 17/12/1985, 07/01/1995, 11/01/1995, 14/02/1995,
> ..., 19/12/1995. In this case, data is defined every 10 years (date format is DD/MM/YYYY).
> You can think of `LinearPeriodFillingInterpolation` as first filling the years in between by
> performing linear interpolation across years, then performs a second linear interpolation
> for the specific day (in actuality, the order of operations is reversed). In this example,
> suppose first we want to evaluate the input on 04/01/1985. Since we have data for 1985, we
> simply perform linear interpolation between 03/01/1985 and 05/01/1985. Suppose now, we want
> to evaluate on 08/01/1987. We do not have data for the year 1987, so, we perform three
> linear interpolations, first to evaluate the data on 08/01/1985 and 08/01/1995 (using the
> neighboring points), and then between the resulting two values.
> 
> `extrapolation_bc` specifies how to deal with out of boundary values. The default value is
> `Throw`, meaning that extrapolation is not allowed.
> 
> When is this interpolation method useful?
> ==========================================
> 
> This interpolation method is useful when certain variations are more important than others.
> For example, suppose you have hourly data defined once a month, on the 15th. A naive
> interpolation method would interpolate use data from 15 Jan at 23.00 and 15 Feb at 00.00 to
> interpolate on any day/hour of the days in between 15 Jan and 15 Feb. However, this would be
> remove the diurnal cycle. `LinearPeriodFillingInterpolation` solves this problem.
> 
> Implementation details
> ======================
> 
> How is `LinearPeriodFillingInterpolation` actually implemented?
> 
> First, let get a high level perspective. Let us continue with the example above.
> 
> There are two cases:
> 
> First, the target date falls within other dates that are boundary dates for the two
> neighboring years. An example is 08/01/1986 because both 08/01/1985 and 08/01/1995 have two
> neighboring dates (05/01/1985, 11/02/1985 and 07/01/1995, 11/01/1995). This is the easy
> case: we interpolate 05/01/1985 and 11/02/1985 to get 08/01/1985, and 07/01/1995 and
> 11/01/1995 to get 08/01/1995, and interpolate the two results to get 08/01/1986.
> 
> Second, the other cases, when the target date does not have two boundary dates defined in
> 1995. An example is 04/01/1986 because 07/01/1995 is the earliest date available on that
> year. We have to be careful with this interpolation because we do not want to mix December
> 1985 with January 1995. In this case, we perform another interpolation to obtain a date in
> December 1994 and reduce to the first case using this date as second half of the
> interpolation bracket. To find what date to interpolate to, we walk the available dates
> backwards in time at find the first date that we can interpolate because it falls into the
> first case. In this example, it is 11/02/1994 (which we can obtain because we have data for
> 11/12/1985 and we can interpolate data on 11/12/1995).

Closes #74

### DIFF
--- a/ext/time_varying_inputs_linearperiodfilling.jl
+++ b/ext/time_varying_inputs_linearperiodfilling.jl
@@ -1,0 +1,375 @@
+# This file is included in TimeVaryingInputsExt.jl
+
+"""
+    _period_difference(date_left, date_right, period::Dates.DatePeriod)
+
+Return the difference in periods (in `Dates.Period`) from `date_right` to `date_left`.
+
+For example, if `date_left` is 18/08/1995, `date_right` 17/01/1997, and `period = Year(1)`,
+the difference is 2 years.
+
+This difference can also be used move a date in `period_right` to `period_left` (or viceversa).
+
+In the example above, 18/08/1995 + 2 years is in `period_right` (or 17/01/1997 - 2 years is
+in `period_left`).
+"""
+function _period_difference(date_left, date_right, period::Dates.DatePeriod)
+    return beginningofperiod(date_right, period) -
+           beginningofperiod(date_left, period)
+end
+
+"""
+    _extract_period(date, period::Dates.DatePeriod)
+
+Return the beginning of the period with duration specified by `period`.
+
+This function is here mostly for clarity (because extracting period clearer in intent than beginning
+of period).
+
+Example
+========
+
+# TODO This jldoctest would require loading the extension...
+
+```
+julia> _extract_period(Date(1993, 8, 18), Year(1))
+DateTime(1993, 1, 1)
+
+julia> _extract_period(Date(1993, 8, 18), Month(1))
+DateTime(1993, 8, 1)
+```
+"""
+function _extract_period(date, period::Dates.DatePeriod)
+    return beginningofperiod(date, period)
+end
+
+"""
+    _neighboring_periods(date, available_periods, period::Dates.DatePeriod)
+
+Return the two periods within `available_periods` that neighbor `date`.
+
+Example
+========
+
+# TODO This jldoctest would require loading the extension...
+
+```
+julia> available_periods = [DateTime(1985, 1, 1), DateTime(1995, 1, 1), DateTime(2005, 1, 1)];
+
+julia> _neighboring_periods([DateTime(1993, 8, 18)], available_periods, Year(1))
+(DateTime(1985, 1, 1), DateTime(1995, 1, 1))
+
+julia> _extract_period(Date(1993, 8, 18), Month(1))
+DateTime(1993, 8, 1)
+```
+"""
+function _neighboring_periods(date, available_periods, period::Dates.DatePeriod)
+    index_period_right =
+        findfirst(d -> d >= beginningofperiod(date, period), available_periods)
+    index_period_left = index_period_right - 1
+    return available_periods[index_period_left],
+    available_periods[index_period_right]
+end
+
+"""
+    _move_date_to_period(date, target_period, period::Dates.DatePeriod)
+
+Return a new date so that it is the equivalent date in the `target_period`
+
+Example
+========
+
+# TODO This jldoctest would require loading the extension...
+
+```
+julia> _move_date_to_period(DateTime(1987, 3, 7), DateTime(1985, 1, 1), Year(1))
+DateTime(1985, 3, 7)
+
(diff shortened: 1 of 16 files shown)


ext/time_varying_inputs_linearperiodfilling.jl
```diff
@@ -0,0 +1,375 @@
...
+
+Check if `date` is between `left` and `right`.
+
+Example
+========
+
+# TODO This jldoctest would require loading the extension...
+
+```
+julia> _date_in_range(DateTime(1993, 19, 11), DateTime(1993, 1, 11), DateTime(1993, 12, 11))
```
szy21, line 109:
```suggestion
julia> _date_in_range(DateTime(1993, 10, 11), DateTime(1993, 1, 11), DateTime(1993, 12, 11))
```
(typo?)

ext/time_varying_inputs_linearperiodfilling.jl
```diff
@@ -0,0 +1,375 @@
...
+
+Example
+========
+
+# TODO This jldoctest would require loading the extension...
+
+```
+julia> dates_period_left = [DateTime(1985, 1, 15), DateTime(1985, 2, 17), DateTime(1985, 12, 11)];
+
+julia> dates_period_right = [DateTime(1995, 1, 8), DateTime(199h5, 2, 18), DateTime(1995, 12, 18)];
```
szy21, line 140:
```suggestion
julia> dates_period_right = [DateTime(1995, 1, 8), DateTime(1995, 2, 18), DateTime(1995, 12, 18)];
```

ext/time_varying_inputs_linearperiodfilling.jl
```diff
@@ -0,0 +1,375 @@
...
+julia> _date_in_range(DateTime(1993, 1, 1), DateTime(1993, 1, 11), DateTime(1993, 12, 11))
+false
+```
+"""
+function _date_in_range(date, left, right)
+    return left <= date <= right
+end
+
+"""
+    _interpolable_range(dates_period_left, dates_period_left)
```
szy21, line 121:
```suggestion
    _interpolable_range(dates_period_left, dates_period_left, period_left)
```

ext/time_varying_inputs_linearperiodfilling.jl
szy21: Also where is dates_period_right used in the docstring?

src/TimeVaryingInputs.jl
```diff
@@ -225,4 +226,83 @@ function LinearInterpolation()
...
+11/01/1995 to get 08/01/1995, and interpolate the two results to get 08/01/1986.
+
+Second, the other cases, when the target date does not have two boundary dates defined in
+1995. An example is 04/01/1986 because 07/01/1995 is the earliest date available on that
+year. We have to be careful with this interpolation because we do not want to mix December
+1985 with January 1995. In this case, we perform another interpolation to obtain a date in
+December 1994 and reduce to the first case using this date as second half of the
+interpolation bracket. To find what date to interpolate to, we walk the available dates
+backwards in time at find the first date that we can interpolate because it falls into the
+first case. In this example, it is 11/02/1994 (which we can obtain because we have data for
```
szy21, line 284:
```suggestion
first case. In this example, it is 11/12/1994 (which we can obtain because we have data for
```

szy21 (approved):
Thank you!

ext/time_varying_inputs_linearperiodfilling.jl
Sbozzolo: Ah yeah, line 144 should have used `dates_period_right`, I fixed it

PR #87, opened 2024-08-07
Fix computing date offset in TimeVaryingInputs

Sbozzolo:
`endofperiod` returns the last second, not millisecond


kmdeck (approved):
looks good/makes sense!

PR #97, opened 2024-08-15
Add Readme badges

Sbozzolo:
image


AlexisRenchon (approved):
You could add downloads too like we have for ClimaLand (if you want... Of course that's not essential)

PR #101, opened 2024-08-21
Improve handling of floatin gpoint errors in DataHandling times

### LINKED ISSUE #99: Error in time in TimeVaryingInputs likely due to floating point error
TimeVaryingInputs sometimes looks for a DateTime slightly different from the one in the dataset, which causes errors. See [here](https://buildkite.com/clima/climacoupler-amip/builds/14#0191704d-9734-4faf-aaff-37f90a3f03db/165-715) for example. The dataset contains 2010-07-17T14:11:30.422 but it is looking for 2010-07-17T14:11:30.421

Sbozzolo:
This bug was found while working with ozone data. Ozone data is defined at non-clean times, eg 2010-07-17T14:11:30.422, and was leading to failures.

The reason for failures are due to moving between times and dates, with times being represented as floating points.

For Float64, times of order of O(few) years (~ 1e7-1e8 seconds) have errors of O(1-100) nanoseconds. As a result, dates like
2010-07-17T14:11:30.422 can be misresolved to 2010-07-17T14:11:30.421 or 2010-07-17T14:11:30.423 when computed from a time. This results in errors because data that is expected to be found is not found.

As a workaround, this commit truncates processing to Milliseconds (because Milliseconds is the time resolution in DateTime) and adds a round operation to recover that resolution.

It is still possible for this workaround to fail: if the floating point error is very large, the round operation might round in the wrong direction.

Closes #99

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,12 @@
 ClimaUtilities.jl Release Notes
 ===============================
 
+v0.1.14
+-------
+
+- Reduce floating point errors in times for DataHandling. PR
+  [#101](https://github.com/CliMA/ClimaUtilities.jl/pull/101)
+
 v0.1.13
 -------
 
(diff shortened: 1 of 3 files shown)


szy21 (approved):
Thanks!

PR #109, opened 2024-09-10
Remove t_start, rename reference_date->start_date

Sbozzolo:
`t_start` and `reference_date` were added following an incorrect assumption (that restarts would restart at t = 0). The names were taken from ClimaLand/ClimaCouper, where these variables were used slightly differently (`reference_date` might have been the first date available in a file, and `t_start` allowed to run at at different point in that file).

Passing Land build: https://buildkite.com/clima/climaland-ci/builds/5083

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,30 +3,40 @@ ClimaUtilities.jl Release Notes
 
 main
 -------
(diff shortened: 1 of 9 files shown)


kmdeck (approved):
I think this looks good, thanks!

PR #116, opened 2024-10-09
Reduce allocations due to isequispaced

Sbozzolo:
`Utils.isequispaced` is now more efficient: it fails fast and does not allocate as much. More redundant allocations due to `Utils.isequispaced` were fixed.

### DIFF
--- a/src/Utils.jl
+++ b/src/Utils.jl
@@ -97,8 +97,13 @@ function isequispaced(
     tol = eltype(v) <: AbstractFloat ? sqrt(eps(eltype(v))) : eps(),
 )
     length(v) < 2 && return true
-    diffs = diff(v)
-    return all(abs(d - diffs[begin]) < tol for d in diffs)
+    dv = v[begin + 1] - v[begin]
+    @inbounds for i in 2:length(v)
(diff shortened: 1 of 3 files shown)


src/Utils.jl
```diff
@@ -97,8 +97,13 @@ function isequispaced(
     tol = eltype(v) <: AbstractFloat ? sqrt(eps(eltype(v))) : eps(),
 )
     length(v) < 2 && return true
-    diffs = diff(v)
-    return all(abs(d - diffs[begin]) < tol for d in diffs)
+    dv = v[begin + 1] - v[begin]
+    @inbounds for i in 2:length(v)
```
kmdeck, line 101: Is it true that this now allocates less because you loop over the array `v`, but only check element by element (allocating scalars), while previously you had to allocate an array all at once (`diffs`)?

kmdeck (approved):
looks good! Thank you!

src/Utils.jl
Sbozzolo: > Is it true that this now allocates less because you loop over the array `v`, but only check element by element (allocating scalars), while previously you had to allocate an array all at once (`diffs`)?

Yes, exactly

PR #83, opened 2024-07-31
Add read!

Sbozzolo:
This commit adds a new in-place function `read!` to read data from files and uses this new method to reduce allocations in regridding.

Closes #81

PR #119, opened 2024-10-18
Add read! function

### LINKED ISSUE #81: Add `FileReader.read!`
Add a method to use a pre-allocated area of memory for data read from NetCDF files.

juliasloan25:
## Purpose 
Adds a new function `read!` that reads in data to a preallocated space in the DataHandler, as opposed to allocating at each call as the existing `read` does. This requires the addition of a `preallocated_read_data` cache to the `DataHandler` object. 

It also requires that `read` returns a copy of the data in the case of a cache hit to avoid overwriting data in the cache (this was the cause of the bug in the original implementation).

Note that we expect higher allocations when using `read` due to the added `copy`, but lower allocations using `read!` over `read`. Internally, we now use `read!` now so users should see reduced allocations.

This feature was initially introduced in https://github.com/CliMA/ClimaUtilities.jl/pull/83/, but there was a bug in that implementation so it has been reworked here

closes #81

### DIFF
--- a/ext/NCFileReaderExt.jl
+++ b/ext/NCFileReaderExt.jl
@@ -174,6 +174,14 @@ end
 Read and preprocess the data at the given `date`.
 """
 function FileReaders.read(file_reader::NCFileReader, date::Dates.DateTime)
+    # For cache hits, return a copy to give away ownership of the data (if we were to just
+    # return _cached_reads[date], modifying the return value would modify the private state
+    # of the file reader)
+
+    if haskey(file_reader._cached_reads, date)
+        return copy(file_reader._cached_reads[date])
+    end
+
     # DateTime(0) is the sentinel value for static datasets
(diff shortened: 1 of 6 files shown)


ext/NCFileReaderExt.jl
```diff
@@ -174,6 +174,11 @@ end
 Read and preprocess the data at the given `date`.
 """
 function FileReaders.read(file_reader::NCFileReader, date::Dates.DateTime)
+    # For cache hits, make and return a copy to avoid overwriting data
```
Sbozzolo, line 177:
```suggestion
# For cache hits, return a copy so to give away ownership of the data (if we were to just
# return _cached_reads[date], modifying the return value would modify the private state
# of the file reader)

```

Sbozzolo (approved):

PR #124, opened 2024-11-06
Add ITime, an integer time type

Sbozzolo:
This commit introduces the `ITime` type, a new approach to handling time and dates within CliMA simulations. `ITime` utilizes integer-based representation and operations, offering three advantages:
- Eliminates floating-point errors.
- Provides a trivial way to go from times to dates.
- Provides an abstraction layer over the calendar system, enabling future support for calendars beyond the Gregorian calendar currently used by Dates.

`ITime` is defined by a `counter`, a `period`, and an optional `start_date`. The counter tracks the number of elapsed periods since the start date and the period is customizable (1 second by default). `ITime`s also support fractional counters (currently represented by `Rational`s) to account for stages in the timestepper loop.

`ITime`s support the arithmetic operations (+, -, *, /, div) while maintaining consistency in the units and the start date. This allows users to just specify `start_date` and `period` for one of their `ITime`s (e.g., `t_start`), and everything will be automatically propagated.

Given an `ITime` `t`, one can obtain its time with `seconds(t)` and the date with `date(t)`. I also added `float(t)` to help with the transition.


src/ITime.jl
```diff
@@ -0,0 +1,331 @@
...
+Second / dt)` for Int64.
+
+# Fields
+- `counter::INT`: The number of clock cycles.
+- `period::DT`: The duration of each cycle.
+- `start_date::START_DATE`: An optional start date.
+"""
+struct ITime{
+    INT <: IntegerOrRatio,
+    DT <: Dates.FixedPeriod,
```
glwagner, line 39:
```suggestion
    DT <: Number,
```

src/ITime.jl
```diff
@@ -0,0 +1,331 @@
...
+    START_DATE <: Union{Nothing, Dates.DateTime},
+}
+    counter::INT
+    period::DT
+    start_date::START_DATE
```
glwagner, line 44:
```suggestion
    epoch::START_DATE
```

PR #127, opened 2024-11-15
Add support for reading data in multiple files

Sbozzolo:
`NCFileReader`s can now read multiple files at the same time. The files have to contain temporal data for the given variable and they are aggregated along the time dimension. To use this feature, just pass a vector of file paths to the constructor.

This capability still has to be surfaced to the TimeVaryingInputs.


NEWS.md
```diff
@@ -4,6 +4,13 @@ ClimaUtilities.jl Release Notes
 main
 ------
 
+#### Support reading time data across multiple files. PR [#127](https://github.com/CliMA/ClimaUtilities.jl/pull/127)
+
+`NCFileReader`s can now read multiple files at the same time. The files have to
+contain temporal data for the given variable and they are aggregated along the
+time dimension. To use this feature, just pass a vector of file paths to the
+constructor.
```
kmdeck, line 12: Do the file paths need to be in order? Maybe a point to clarify

docs/src/filereaders.md
```diff
@@ -20,8 +19,10 @@ Future extensions might include:
 > This extension is loaded when loading `NCDatasets`
 
 The only file reader currently implemented is the `NCFileReader`, used to read
-NetCDF files. Each `NCFileReader` is associated to one particular file and
-variable (but multiple `NCFileReader`s can share the same file).
+NetCDF files. Each `NCFileReader` is associated to a collection of files
+(possibly just one) and variable (but multiple `NCFileReader`s can share the
```
kmdeck, line 23: "and one variable"

docs/src/filereaders.md
```diff
@@ -20,8 +19,10 @@ Future extensions might include:
...
+same file). When a `NCFileReader` is constructed with multiple files, the
+various files should contain the time developlment of the given variable.
```
kmdeck, line 25: development

NEWS.md
Sbozzolo: I don't think they need to be in order. I think that everything will work even if the order of dates is random (because we always access one date at the time)

test/file_readers.jl
```diff
@@ -55,6 +55,17 @@ using NCDatasets
...
+
+    # Test times split across multiple files
+    PATHS = [
+        joinpath(@__DIR__, "test_data", "era5_1979_1.0x1.0_lai.nc"),
+        joinpath(@__DIR__, "test_data", "era5_1980_1.0x1.0_lai.nc"),
+    ]
+    NCDataset(PATHS, aggdim = "time") do nc
+        ncreader_agg = FileReaders.NCFileReader(PATHS, "lai_lv")
+        FileReaders.available_dates(ncreader_agg) == nc["time"][:]
+        length(FileReaders.available_dates(ncreader_agg)) == 8
```
kmdeck, line 67: Does 8 make sense? maybe you thinned the data. I thought it was weekly.

NEWS.md
Sbozzolo: Added a note

test/file_readers.jl
Sbozzolo: Yes, they were reduced size. Now, it the full data

ext/NCFileReaderExt.jl
```diff
@@ -64,27 +66,64 @@ end
...
+        # Let's first try to identify the time dimension, if it exists. To do that, we open the
+        # first dataset. We need this to aggregate multiple datasets, if data is split across
+        # multiple files
+        NCDatasets.NCDataset(first(file_paths)) do first_dataset
+            is_time = x -> x == "time" || x == "date" || x == "t"
+            time_dims = filter(is_time, NCDatasets.dimnames(first_dataset))
+            if !isempty(time_dims)
+                aggtime_kwarg = (:aggdim => first(time_dims),)
+            else
+                error("Multiple files given, but no time dimension")
```
kmdeck, line 109: "Multiple files given, but at least one does not have the time dimension "t", "date", or "time"."?

ext/NCFileReaderExt.jl
```diff
@@ -64,27 +66,64 @@ end
...
+            end
+        end
+    end
+
+    # When we have only no time data, we have to pass this as a string
+    file_path_to_ncdataset =
```
kmdeck, line 115: This is not an error because this same code is also used in the case when we have spatially varying but temporally constant data?

kmdeck: so 
"isempty(aggtime_kwarg) ? first(file_paths) " means give me file_paths[1] which is the only element in file_paths

test/file_readers.jl
```diff
@@ -55,6 +55,17 @@ using NCDatasets
...
     end
+
+    # Test times split across multiple files
+    PATHS = [
+        joinpath(@__DIR__, "test_data", "era5_1979_1.0x1.0_lai.nc"),
+        joinpath(@__DIR__, "test_data", "era5_1980_1.0x1.0_lai.nc"),
+    ]
+    NCDataset(PATHS, aggdim = "time") do nc
+        ncreader_agg = FileReaders.NCFileReader(PATHS, "lai_lv")
+        FileReaders.available_dates(ncreader_agg) == nc["time"][:]
```
kmdeck, line 66: Do we need to test that these are also in ascending order?

Sbozzolo: No, it will still work with times in any order

ext/NCFileReaderExt.jl
Sbozzolo: Yes, exactly

ext/NCFileReaderExt.jl
Sbozzolo: Updated

PR #128, opened 2024-11-15
Turn ClimaComms into required dependency

Sbozzolo:
The `OutputPathGenerator` is no longer loaded automatically and is now loaded by loading `ClimaComms`. The motivation behind this change is to provide a more reasonable default to `OutputPathGenerator` (now it is `ClimaComms.context()`). This should improve robustness of `OutputPathGenerator` in MPI settings.


charleskawczynski (commented):
Can we instead just depend on ClimaComms? It's a lightweight package, especially when MPI/CUDA are not loaded.

This way we can safely and more simply use the singleton contexts/ cpu devices as defaults.

Sbozzolo:
Making ClimaComms a dependency would only move the code that I moved to `ext` back to `src`. Given that ClimaComms is already loaded by all our downstream packages, there's no difference in how the module works. So, I'd prefer keeping it ClimaComms a weak dependency and have zero required dependencies in ClimaUtilities, unless that becomes too burdersome. This way, ClimaUtilities can be thought of a zero-cost toolkit that can always be added and imported. 

That said, I wouldn't be opposed to making a dependency if, in the future, there are cases where having ClimaComms as weak dependency leads to unnecessary pain.

charleskawczynski:
> Making ClimaComms a dependency would only move the code that I moved to `ext` back to `src`. Given that ClimaComms is already loaded by all our downstream packages, there's no difference in how the module works. So, I'd prefer keeping it ClimaComms a weak dependency and have zero required dependencies in ClimaUtilities, unless that becomes too burdersome. This way, ClimaUtilities can be thought of a zero-cost toolkit that can always be added and imported.
> 
> That said, I wouldn't be opposed to making a dependency if, in the future, there are cases where having ClimaComms as weak dependency leads to unnecessary pain.

IMO, we should leave it as a direct dependency. Otherwise, we all of the docs like that make MPI-safe-related claims, e.g.,:

> [ClimaArtifacts](https://clima.github.io/ClimaUtilities.jl/dev/climaartifacts/), a module that provides an MPI-safe way to lazily download artifacts and to tag artifacts that are being accessed in a given simulation.

kind of need an asterisk in that they're only valid when ClimaComms (and optionally MPI) is loaded.

One of the main reasons for extensions is to reduce code loading costs, and ClimaComms has reduced its code loading costs by making all of its dependencies extensions, so it doesn't have significant code loading costs.

Making ClimaComms an extension results in more empty function stubs, more code duplication, and complicates the code loading.

For example, `root_or_singleton` basically duplicates `ClimaComms.iamroot` (which can be dispatched between different cases).

If we're really concerned about code loading times with ClimaComms, I'd argue that we should measure the code loading time between what is on main and these changes.

I think we should fix the existing logic we discussed (having default `context::Nothing` methods), and worry about code loading later, if it's an issue.

Sbozzolo:
> kind of need an asterisk in that they're only valid when ClimaComms (and optionally MPI) is loaded.

ClimaArtifacts also works without ClimaComms, but it MPI-safe only with ClimaComms. I updated that line in the docs.

> I think we should fix the existing logic we discussed (having default context::Nothing methods), and worry about code loading later, if it's an issue.

Okay, this is fixed in this PR. We can worry about code loading (ie, whether to make ClimaComms a direct dependency or not) later.

Sbozzolo:
Okay, fair. I turned `ClimaComms` into a required dependency.

src/OutputPathGenerator.jl
```diff
@@ -251,12 +250,12 @@ function generate_output_path(::ActiveLinkStyle, output_path; context = nothing)
             end
         end
     end
+    new_output_folder = bcast(context, new_output_folder)
```
charleskawczynski, line 253: Do we need the `maybe_wait_filesystem` after this?

src/OutputPathGenerator.jl
```diff
@@ -168,70 +168,69 @@ end
...
-    # We need the path_separator as a char to use it in rstrip
-    path_separator_char = path_separator_str[1]
-    output_path = rstrip(output_path, path_separator_char) * path_separator_char
-
-    # Create folder if it does not exist
-    if root_or_singleton(context)
+    # Only root needs to do file-system operations, we will broadcast the result
+    # to the other MPI processes to the new_output_folder variable after
+    # everything is done.
+    new_output_folder = ""
```
charleskawczynski, line 174:
```suggestion
    new_output_folder = "" # needed for ClimaComms bcast, but is ignored
```

src/OutputPathGenerator.jl
Sbozzolo: Yes, because broadcasting the folder name doesn't guarantee that the folder exists on all the nodes

src/OutputPathGenerator.jl
```diff
@@ -168,70 +168,69 @@ end
...
+                filter(x -> !isnothing(match(name_rx, x)), readdir(output_path))
+            if length(existing_outputs) > 0
+                @warn "$output_path already contains some output data, but no active link"
+                latest_output = first(sort(existing_outputs, rev = true))
+                counter_str = match(name_rx, latest_output)
+                counter = parse(Int, counter_str[1])
+                next_counter = counter + 1
+                @warn "Restarting counter from $next_counter"
+            else
+                # This is our first counter
```
charleskawczynski, line 222:
```suggestion
                # This is our first output folder, initialize counter
```

charleskawczynski (approved):
Overall looks good. I think this is a lot cleaner. Thanks @Sbozzolo!

src/OutputPathGenerator.jl
charleskawczynski: Right, looks good

PR #132, opened 2024-11-26
Expose support for multiple files to DataHandler

### LINKED ISSUE #126: Support reading time data across multiple files
If data is split across files across the time dimension, this can be read in a single go with `aggdim`. We should add support for that

Sbozzolo:
This commit surfaces the new capability introduced recently in NCFileReaders to read multiple files as if they were one.

Closes #126

### DIFF
--- a/docs/src/datahandling.md
+++ b/docs/src/datahandling.md
@@ -9,7 +9,7 @@ This is no trivial task. Among the challenges:
 - IO can be very expensive,
 - CPU/GPU communication can be a bottleneck.
 
-The `DataHandling` takes the divide and conquer approach: the various core tasks
+The `DataHandling` takes the divide-and-conquer approach: the various core tasks
 and features and split into other independent modules (chiefly
 [`FileReaders`](@ref), and [`Regridders`](@ref)). Such modules can be developed,
 tested, and extended independently (as long as they maintain a consistent
@@ -17,9 +17,10 @@ interface). For instance, if need arises, the `DataHandler` can be used (almost)
 directly to process files with a different format from NetCDF.
 
 The key struct in `DataHandling` is the `DataHandler`. The `DataHandler`
-contains one or more `FileReader`(s), a `Regridder`, and other metadata necessary to perform
-its operations (e.g., target `ClimaCore.Space`). The `DataHandler` can be used
-for static or temporal data, and exposes the following key functions:
+contains one or more `FileReader`(s), a `Regridder`, and other metadata
+necessary to perform its operations (e.g., target `ClimaCore.Space`). The
+`DataHandler` can be used for static or temporal data, and exposes the following
+key functions:
 - `regridded_snapshot(time)`: to obtain the regridded field at the given `time`.
                     `time` has to be available in the data.
 - `available_times` (`available_dates`): to list all the `times` (`dates`) over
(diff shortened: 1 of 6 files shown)


docs/src/datahandling.md
```diff
@@ -108,6 +121,10 @@ function linear_interpolation(data_handler, time)
 end
 ```
 
+If, for example, the data was split across multiple files named `era5_1980.nc`,
+`era5_1981.nc`, ... (e.g., each file containing one year), we could directly
+pass the list to `DataHandler`, which knows how to combine them.
```
juliasloan25, line 126: It might be good to clarify that in this case the list would be passed as the first argument to DataHandler

docs/src/inputs.md
```diff
@@ -88,6 +88,26 @@ data_handler = TimeVaryingInputs.TimeVaryingInput("era5_example.nc",
 The same arguments (excluding `start_date`) could be passed to a
 `SpaceVaryingInput` to compose multiple input variables with that type.
 
+#### Example: Data split across multiple NetCDF files
+
+Often, large datasets come chunked, meaning that the data is split across
+multiple files. `TimeVaryingInput`s know to combine data across multiple files
```
juliasloan25, line 94: We should specify that this is intended for data that's split across the time dimension

ext/DataHandlingExt.jl
```diff
@@ -93,9 +93,61 @@ struct DataHandler{
...
+function _check_file_paths_varnames(
+    file_paths,
+    varnames,
+    regridder_type,
+    compose_function,
+)
+    # Verify that the number of file paths and variable names are consistent
+    if length(varnames) == 1
+        # When we have one varname, this means that all the files contain the same variable
+        # In this case, the `compose_function` has to be identify
```
juliasloan25, line 112:
```suggestion
        # In this case, the `compose_function` has to be identity
```

ext/DataHandlingExt.jl
```diff
@@ -93,9 +93,61 @@ struct DataHandler{
...
+        compose_function == identity || error(
+            "`compose_function` must be identity when using a single input variable",
+        )
+
+        # Multiple files are not not supported by TempestRemapper
+        (length(file_paths) > 1 && regridder_type == :TempestRegridder) &&
+            error("TempestRegridder does not support multiple input files")
+    else
+        # We have multiple variables
+        # This is not supported by TempestRemapper
```
juliasloan25, line 122:
```suggestion
        # This is not supported by TempestRegridder
```

ext/DataHandlingExt.jl
```diff
@@ -114,7 +172,21 @@ Creating this object results in the file being accessed (to preallocate some mem
 Positional arguments
 =====================
 
-- `file_paths`: Paths of the NetCDF file(s) that contain the input data.
+- `file_paths`: Paths of the NetCDF file(s) that contain the input data. There are three
+  options for `file_paths`:
+  - It is a string that points to a single NetCDF file.
+  - It is a list that points to a multiple NetCDF file. In this case, we support two modes:
```
juliasloan25, line 178:
```suggestion
  - It is a list that points to multiple NetCDF files. In this case, we support two modes:
```

ext/DataHandlingExt.jl
```diff
@@ -114,7 +172,21 @@ Creating this object results in the file being accessed (to preallocate some mem
...
+    1. `varnames` is a string or an iterable with one element only. We take this to mean
+       that the NetCDF files contain the time development of that variable.
+    2. `varnames` has multiple elements. We take this to mean that the NetCDF files contain
+       different variables. In this case, the order has to match the order in the
+       `file_paths` and has to match the order of arguments in `compose_function`.
+  - It is a list of lists of paths to NetCDF files, where the inner list identifies temporal
+    chunks of a given variable, and the outer list identifies different variables
+    (supporting the mode where different variables live in different files and their time
+    development is split across multiple files). This is currently not implemented, but can
+    be implemented if you need it. Please open an issue if this is the case.
```
juliasloan25, line 188: It seems a bit misleading to list this as one of the 3 available options for `file_paths`, but then to say it isn't implemented yet. If you're planning to add it soon though I think this is fine

juliasloan25 (approved):
Thank you!

docs/src/inputs.md
```diff
@@ -88,6 +88,26 @@ data_handler = TimeVaryingInputs.TimeVaryingInput("era5_example.nc",
...
+as it were provided in a single file. To do use this feature, just pass the list
+of file paths. While it is not required for the files to be in order, it is good
+practice to pass them in ascending order.
```
kmdeck, line 97: "ascending order by time"

ext/DataHandlingExt.jl
```diff
@@ -93,9 +93,61 @@ struct DataHandler{
...
+    file_paths,
+    varnames,
+    regridder_type,
+    compose_function,
+)
+    # Verify that the number of file paths and variable names are consistent
+    if length(varnames) == 1
+        # When we have one varname, this means that all the files contain the same variable
+        # In this case, the `compose_function` has to be identify
+        compose_function == identity || error(
```
kmdeck, line 113: Is this true? Dont we support unit conversions? e.g. evap -> evap/1000 to convert from a mass flux to volume flux of water

ext/DataHandlingExt.jl
Sbozzolo: Now, I implemented this

ext/DataHandlingExt.jl
Sbozzolo: I mean, I guess you could use the compose function to do unit conversion, but what you are actually thinking of is the preprocess_func.

The difference is this: the compose function operates on the regridded files, while the preprocessing function on the raw data. I removed this error because indeed one could use the compose function to do data processing

docs/src/datahandling.md
```diff
@@ -66,6 +67,50 @@ are composed.
...
+    compose_function = (x, y) -> x + y,
+)
+```
+
+In this case, `DataHandler` will incorrectly assume that `lai_hv` is contained
+in `era1980.nc`, and `lai_lv` is in `era1980.nc`. Instead, construct the
+`data_handler` by passing a list of list 
+```julia
+files = ["era1980.nc", "era1981.nc"] 
+data_handler = DataHandling.DataHandler(
```
kmdeck, line 104: good explanation!

ext/DataHandlingExt.jl
```diff
@@ -138,13 +189,15 @@ everything more type stable.)
                       It can be a NamedTuple, or a Dictionary that maps Symbols to values.
 - `file_reader_kwargs`: Additional keywords to be passed to the constructor of the file reader.
                         It can be a NamedTuple, or a Dictionary that maps Symbols to values.
-- `compose_function`: Function to combine multiple input variables into a single data variable.
-                    The default, to be used in the case of one input variable, is the identity.
-                    Note that the order of `varnames` must match the argument order of `compose_function`.
+- `compose_function`: Function to combine multiple input variables into a single data
```
kmdeck, line 192: unrelated to this PR but, we could also note that, since the variables will be matrices or similar, your compose function must broadcast correctly (this tripped me up initially)

kmdeck (approved):
looks good!  thanks!

ext/DataHandlingExt.jl
Sbozzolo: Added!

PR #133, opened 2024-11-26
Reduce default max cache size for DataHandlers

Sbozzolo:
This should reduce the memory footprint for long simulations

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -43,6 +43,12 @@ to learn more about this.
 
 This capability is only available for the `InterpolationsRegridder`.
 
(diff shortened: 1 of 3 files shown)


szy21 (approved):

juliasloan25:
Why not keep the default value and just pass in a smaller value for the high resolution runs we need it for?

Sbozzolo:
> Why not keep the default value and just pass in a smaller value for the high resolution runs we need it for?

I think that as long as the DataHandler is used in a TimeVaryingInput that is marching forward, we will only need up to two caches, so 2 is a better default everywhere. Also, changing it to a smaller value would require changing every instance of TimeVaryingInput in ClimaAtmos (and elsewhere to)

juliasloan25 (approved):

PR #134, opened 2024-11-27
Fix notes in docs

### LINKED ISSUE #129: Clean `note`s in the documentation
Instead of `> :note:`, it should be `!!! note`

Sbozzolo:
The correct notation is
```
!!! note
```

Not what I had before

Closes #129


juliasloan25 (approved):

imreddyTeja (commented):
This is already merged, but `!!!note` should be `!!! note`

PR #136, opened 2024-12-02
Make DataHandler automatically reverse dimensions

### LINKED ISSUE #131: InterpolationsRegridder does not produce any warning when the dimensions are not propertly sorted
`Interpolations.jl` only supports interpolating with monothonically increasing dimensions. When this is not true, the error is difficult to understand and trace back. We can catch this in InterpolationsRegridder upon initialization and return a more meaninfgul errror.

imreddyTeja:
When a DataHandler is created with the constructor and the regridder type is InterpolationsRegridder, the constructor checks each dimension. If the dimension is monotonically decreasing, the dimension will be reversed before each regrid. If the dimension is not monotonically increasing or decreasing, then an error that states the problematic dimension is thrown.

The dimension names of the read file are added to the NCFileReader struct. This is used to give the name of the problematic dimension if an error is thrown. This could be removed if the error message just gives the index of the problematic dimension.

The InterpolationsRegridder now has a dim_trasforms field. dim_transform is expected to be a tuple of functions with the same length as the number of spatial dimensions. When regridding is performed, the transforms are applied to each dimension, and then to the data in each dimension.

## Purpose 
Closes #131 

## Content
- informative error thrown when a dimension is not monotonically increasing or decreasing
- automatically deals with monotonically decreasing dimensions
- allows for other dimension correcting functions to applied if specified

----


ext/InterpolationsRegridderExt.jl
```diff
@@ -81,8 +104,36 @@ This function is allocating.
...
+            end
+        end
+        for i in 1:length(dimensions_FT[2])
+            for j in 1:length(dimensions_FT[3])
+                data[:, i, j] .= regridder.dim_transforms[1](data[:, i, j])
+            end
+        end
+    else
+        error("Only 2D and 3D data is supported")
+    end
```
Sbozzolo, line 136: This modifies `data` inplace, which might be undersirable. Note that the function `reverse` takes a keyword argument `dims`, so you could use directly that.

ext/InterpolationsRegridderExt.jl
```diff
@@ -69,7 +79,20 @@ function Regridders.InterpolationsRegridder(
...
     end
 
-    return InterpolationsRegridder(target_space, coordinates, extrapolation_bc)
+    if isnothing(dim_transforms)
+        if eltype(coordinates) <: ClimaCore.Geometry.LatLongPoint
+            dim_transforms = (identity, identity)
+        elseif eltype(coordinates) <: ClimaCore.Geometry.LatLongZPoint
+            dim_transforms = (identity, identity, identity)
+        end
+    end
```
Sbozzolo, line 88: I think you can use the length of `coordinates` for this. Something like
```
dim_transforms = ntuple(_ -> identity, length(coordinates))
```

ext/InterpolationsRegridderExt.jl
```diff
@@ -81,11 +104,48 @@ This function is allocating.
...
+        end
+        for i in 1:length(dimensions_FT[2])
+            for j in 1:length(dimensions_FT[3])
+                data_transformed[:, i, j] .=
+                    regridder.dim_transforms[1](data_transformed[:, i, j])
+            end
+        end
+    else
+        error("Only 2D and 3D data is supported")
+    end
```
Sbozzolo, line 141: I am wondering if we should just sticking at reversing the dimension (that we can do very easily) instead of supporting arbitrary transformations that leads to harder to read/maintain code

ext/InterpolationsRegridderExt.jl
imreddyTeja: That does mnot work because the length of any field is 1. See [here](https://github.com/CliMA/ClimaCore.jl/blob/340603b306e5fa54a1fe6db5bdfff6b5e2aadfb5/src/Fields/Fields.jl#L178). I couldn't figure out any easy way to get the number of dimensions from a field.

ext/DataHandlingExt.jl
```diff
@@ -323,7 +329,14 @@ function DataHandling.DataHandler(
         regridder_args =
             (target_space, first(varnames), first(first(file_paths)))
     elseif regridder_type == :InterpolationsRegridder
+        # Check if the dimensions are monotonically increasing or decreasing
+        # if monotonically decreasing, we need to reverse the data before regridding
+        dim_decreasing = map(dimensions, dim_names) do dim, dim_name
+            @assert issorted(dim) || issorted(reverse(dim)) "dimension $dim_name is not monotonically increasing or decreasing"
+            issorted(dim) ? false : true
+        end
```
Sbozzolo, line 337:
```suggestion
        dim_decreasing = map(dimensions, dim_names) do dim, dim_name
            @assert issorted(dim) || issorted(dim, rev = true) "dimension $dim_name is neither monotonically increasing nor decreasing"
            issorted(dim)
        end
```

ext/InterpolationsRegridderExt.jl
```diff
@@ -22,6 +23,9 @@ struct InterpolationsRegridder{
 
     """Tuple of extrapolation conditions as accepted by Interpolations.jl"""
     extrapolation_bc::BC
+
+    """Tuple of booleans signifying if the dimension is monotonically decreasing"""
```
Sbozzolo, line 27:
```suggestion
    """Tuple of booleans signifying if the dimension is monotonically decreasing. True for dimensions that are monotonically increasing, false for dimensions that are monotonically decreasing."""
```

ext/InterpolationsRegridderExt.jl
```diff
@@ -81,11 +97,20 @@ This function is allocating.
...
 function Regridders.regrid(regridder::InterpolationsRegridder, data, dimensions)
     FT = ClimaCore.Spaces.undertype(regridder.target_space)
-    dimensions_FT = map(d -> FT.(d), dimensions)
-
+    dimensions_FT = map(dimensions, regridder.dim_decreasing) do dim, decreasing
+        decreasing ? reverse(FT.(dim)) : FT.(dim)
+    end
+    decreasing_indices =
+        Tuple([i for (i, d) in enumerate(regridder.dim_decreasing) if d])
+    # apply the specified transformation to each dimension of the data
```
Sbozzolo, line 105:
```suggestion
```

ext/InterpolationsRegridderExt.jl
```diff
@@ -81,11 +97,20 @@ This function is allocating.
...
+    data_transformed = reverse(data, dims = decreasing_indices)
```
Sbozzolo, line 106: Could you check if `reverse` allocates when nothing should be reversed? If that's the case, we should add handle it in such a way that when nothing has to be reversed, `reverse` is not called.

test/regridders.jl
```diff
@@ -29,6 +29,107 @@ include("TestTools.jl")
...
+        data_lat2D_reversed,
+        dimensions2D_reversed,
+    )
+    regridded_lon_reversed =
+        Regridders.regrid(reg_horz_reversed, data_lon2D, dimensions2D_reversed)
+    @test regridded_lat_reversed == regridded_lat
+    @test regridded_lon_reversed == regridded_lon
+
+    # create one regridder with no transformations to dimensions needed
+    # create another regridder that reverses the second and third dimensions
```
Sbozzolo, line 105:
```suggestion
    # Create one regridder with no transformations to dimensions needed
    # Create another regridder that reverses the second and third dimensions
```

Sbozzolo (approved):

PR #137, opened 2024-12-04
Add OnlineLogging

Sbozzolo:
`OnlineLogging` is a new module (mostly based on ClimaAtmos `WallTimeEstimate`) that contains tools to provide informative messages while a simulation is running.

At the moment, `OnlineLogging` only contains `WallTimeInfo`, to report on timing information.

This new module will allow ClimaLand and ClimaCoupler to add some progress information, and remove some entries from the atmos cache.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,8 @@ ClimaUtilities.jl
   [`DataHandling`](https://clima.github.io/ClimaUtilities.jl/dev/datahandling/),
   and [`Regridders`](https://clima.github.io/ClimaUtilities.jl/dev/regridders/)
   to process input data and remap it onto the simulation grid.
+- [`OnlineLogging`](https://clima.github.io/ClimaUtilities.jl/dev/onlinelogging/)
+  to add output information while a simulation is running.
 - [`OutputPathGenerator`](https://clima.github.io/ClimaUtilities.jl/dev/outputpathgenerator/)
   to prepare the output directory structure of a simulation.
 - [`TimeManager`](https://clima.github.io/ClimaUtilities.jl/dev/timemanager/) to
--- a/docs/src/onlinelogging.md
+++ b/docs/src/onlinelogging.md
@@ -1,0 +1,89 @@
+# Online Simulation Progress Reporting
+
+The `OnlineLogging` module provides tools to monitor and report the simulation
+progress and other information of simulations.
+
+Currently, the only feature implemented is timing report, to print information
+about current step, simulation time, and average performance. With
+`report_walltime`, you will see progress information like the following
+```
+┌ Info: Progress
+│   simulation_time = "2 seconds"
+│   n_steps_completed = 20
+│   wall_time_per_step = "10 milliseconds, 100 microseconds"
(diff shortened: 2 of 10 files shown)


README.md
```diff
@@ -24,6 +24,8 @@ ClimaUtilities.jl
   [`DataHandling`](https://clima.github.io/ClimaUtilities.jl/dev/datahandling/),
   and [`Regridders`](https://clima.github.io/ClimaUtilities.jl/dev/regridders/)
   to process input data and remap it onto the simulation grid.
+- [`OnlineLogging`](https://clima.github.io/ClimaUtilities.jl/dev/onlinelogging/)
+  add output information while a simulation is running.
```
juliasloan25, line 28:
```suggestion
  to add output information while a simulation is running.
```

docs/src/onlinelogging.md
```diff
@@ -0,0 +1,87 @@
...
+└   estimated_finish_date = 2024-12-03T16:01:14.660
+```
+
+## `WallTimeInfo` and `report_walltime`
+
+`WallTimeInfo` is a struct that holds information about wall time (the time you
+see on your watch, not the simulation time) and that can be used to report
+timing information with [`report_walltime`](@ref).
+
+`WallTimeInfo` keeps track and accumulate how much time has elapsed since the
```
juliasloan25, line 29:
```suggestion
`WallTimeInfo` keeps track and accumulates how much time has elapsed since the
```

docs/src/onlinelogging.md
```diff
@@ -0,0 +1,87 @@
...
+last time it was updated. In this, `WallTimeInfo` tries to automatically remove
+the compilation overhead that your simulation might run into in the first step.
```
juliasloan25, line 31: How is this done? Does it run everything twice?

src/OnlineLogging.jl
```diff
@@ -0,0 +1,201 @@
...
+        n_steps wall_time_per_step = wall_time_ave_per_step_str wall_time_total =
+        wall_time_total wall_time_remaining = wall_time_remaining_str wall_time_spent =
+        wall_time_spent percent_complete = "$percent_complete%" estimated_sypd =
+        sypd_estimate date_now = Dates.now() estimated_finish_date =
+        estimated_finish_date
+
+    return nothing
+end
+
+# TODO: Clean this up
```
juliasloan25, line 186: Could you remove these TODOs before merging?

Sbozzolo: I was planning on tackling this as part of the integer time PR because I will probably move things around and unify functions

docs/src/onlinelogging.md
Sbozzolo: This should be documented in the comments: we essentially ignore the first two calls and "extrapolate" back once we started having data

juliasloan25: I think it would be good to briefly mention this in the comments so users can understand it without looking into the source code

juliasloan25 (approved):

PR #140, opened 2024-12-11
Tr/minor tempest regridder changes

imreddyTeja:
## Purpose 

- ~~Make TempestRegridder support multiple vars (maybe)~~ This is likely not needed. It should be its own PR if we ever decide it is needed, and the implementation/interface requires planning to reflect the design of InterpolatonsRegridder
- ~~Make outfile_root a kwarg for TempestRegridder~~ Just remove it.
- ~~improve on read_available_dates~~
- Add tests for TempestRegridder and read_available_dates
- Replace `yyyymmdd_to_datetime` with Dates built in parsing

## Content

----

### DIFF
--- a/ext/TempestRegridderExt.jl
+++ b/ext/TempestRegridderExt.jl
@@ -33,12 +33,10 @@ struct TempestRegridder{
     SPACE <: ClimaCore.Spaces.AbstractSpace,
     STR1 <: AbstractString,
     STR2 <: AbstractString,
-    STR3 <: AbstractString,
 } <: Regridders.AbstractRegridder
     target_space::SPACE
     regrid_dir::STR1
-    outfile_root::STR2
-    varname::STR3
+    varname::STR2
     mono::Bool
 end
 
@@ -75,7 +73,6 @@ function Regridders.TempestRegridder(
 )
     space = target_space
     comms_ctx = ClimaComms.context(space)
-    outfile_root = varname
 
     if ClimaComms.iamroot(comms_ctx)
         @info "Saving TempestRegrid files to $regrid_dir"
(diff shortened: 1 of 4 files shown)


ext/TempestRegridderExt.jl
```diff
@@ -72,10 +72,10 @@ function Regridders.TempestRegridder(
     input_file::AbstractString;
     regrid_dir::AbstractString,
     mono::Bool = true,
+    outfile_root::AbstractString = "cgll",
```
Sbozzolo, line 75: Why is this needed?

imreddyTeja: There is no real need for this to be a kwarg, or `outfile_root` to be changed in general. 

Files saved during Regridding currently have a name structure as 

```
hd_outfile_root * "_" * varname * "_" * string(time) * ".hdf5"
```

but `hd_outfile_root` is automatically set to `varname`, so the path becomes

```
varname * "_" * varname * "_" * string(time) * ".hdf5"
```
, which seems odd to me. The changes/potential changes in this PR came up in [this](https://github.com/CliMA/ClimaCoupler.jl/pull/1109) PR

Sbozzolo: Ah okay! I guess we can directly removed the `hd_outfile_root` then

juliasloan25: I agree it's okay to remove it. It would be good to have an organized naming of regridded files when/if we support regridding a file that contains multiple variables. But that can be accomplished using `regrid_dir` rather than `outfile_root`.

juliasloan25 (approved):

Sbozzolo (approved):
Thank you!

PR #142, opened 2024-12-13
Clean up docs

### LINKED ISSUE #141: Warning while building docs
There are a variety of warnings produced when building the docs. Fixing these should make the docs easier to use. 

Fixing all the warnings likely requires writing docstrings for functions that do not have one. 

Types of Warnings:

```
Warning: duplicate docs

Warning: no docs

Warning: invalid local link

Warning: 6 docstrings not included
```

imreddyTeja:
This commit fixes all warnings in the docs,
except for the warning that the modules' docstrings are not included in the documentation.

The warning for Base.close() is no longer thrown,
but the correct docstrings are not shown in the correct places. This cannot be fixed unless the necessary types are exported from the relevant modules.

Closes #141 

## Content

----

### DIFF
--- a/docs/src/datahandling.md
+++ b/docs/src/datahandling.md
@@ -1,4 +1,4 @@
-# `DataHandling`
+# [`DataHandling`](@id datahandling_module)
 
 The `DataHandling` module is responsible for reading data from files and
 resampling it onto the simulation grid.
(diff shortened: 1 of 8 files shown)


Sbozzolo (approved):
Thank you!

PR #144, opened 2024-12-17
Docs: add example of changing boundary conditions

Sbozzolo:
This was a little hidden.

### DIFF
--- a/docs/src/inputs.md
+++ b/docs/src/inputs.md
@@ -117,11 +117,19 @@ Read more about this feature in the page about [`DataHandler`](@ref datahandling
 default, the `Throw` condition is used, meaning that interpolating onto a point
 that is outside the range of definition of the data is not allowed. Other
 boundary conditions are allowed. With the `Flat` boundary condition, when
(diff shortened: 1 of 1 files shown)


docs/src/inputs.md
```diff
@@ -117,11 +117,19 @@ Read more about this feature in the page about [`DataHandler`](@ref datahandling
 default, the `Throw` condition is used, meaning that interpolating onto a point
 that is outside the range of definition of the data is not allowed. Other
 boundary conditions are allowed. With the `Flat` boundary condition, when
-interpolating outside of the range of definition, return the value of the
-of closest boundary is used instead.
+interpolating outside of the range of definition, return the value of the of
+closest boundary is used instead.
```
imreddyTeja, line 121: This is not something changed in this commit, but "return the value of the of closest boundary is used instead" does not make sense to me.

imreddyTeja (approved):

docs/src/inputs.md
Sbozzolo: Thank you!

PR #147, opened 2025-01-15
Changes the internal behavior of TimeVaryingInputs23D to use dates rather than seconds since start date

### LINKED ISSUE #146: Change TimeVaryingInputs23D to internally use dates instead of seconds from start date
For preparation of ITime, TimeVaryingInputs23D should be updated to internally use dates rather than seconds from start date. Backward compatibility can be maintained for float by converting the float to a date and passing that along.

ph-kev:
closes #146 - This commit changes the internal behavior of TimeVaryingInputs23D to use dates rather than seconds since start date. This will make the transition to ITime a little easier since it is easy to convert a ITime to a date.

The behavior of TimeVaryingInputs23D will only differ when `dt` in milliseconds is an odd number or precision more than one millisecond is required. For the first case, this is because the code computes `dt / 2` in `_time_range_dt_dt_e`. In all other cases, the behavior should be identical as before.

### DIFF
--- a/ext/TimeVaryingInputsExt.jl
+++ b/ext/TimeVaryingInputsExt.jl
@@ -33,12 +33,13 @@ import ClimaUtilities.TimeVaryingInputs: extrapolation_bc
 import ClimaUtilities.DataHandling
 import ClimaUtilities.DataHandling:
     DataHandler,
-    previous_time,
-    next_time,
     regridded_snapshot!,
     available_times,
     available_dates,
-    date_to_time
+    date_to_time,
+    time_to_date,
+    previous_date,
+    next_date
 
 # Ideally, we should be able to split off the analytic part in a different
 # extension, but precompilation stops working when we do so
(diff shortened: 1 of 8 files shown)


ph-kev:
I am running the nightly build locally on the main branch and it is stuck after finishing the DataStructures tests.

ext/TimeVaryingInputsExt.jl
```diff
@@ -198,19 +206,38 @@ function TimeVaryingInputs.evaluate!(
...
+    dest,
+    itp::InterpolatingTimeVaryingInput23D,
+    time::Number,
+    args...;
+    kwargs...,
+)
+    TimeVaryingInputs.evaluate!(
+        dest,
+        itp,
+        Dates.Millisecond(round(Int64, 1000 * time)) +
```
Sbozzolo, line 233: Do you need to specify `Int64`?

Sbozzolo (commented):
Thank you!

I left just one question on the rounding

ext/TimeVaryingInputsExt.jl
ph-kev: I don't think so. I just got rid of it.

PR #148, opened 2025-01-21
Simplify TimeVaryingInputs0D

Sbozzolo:
Recent PRs in `ClimaCore` and
`ClimaComms` (https://github.com/CliMA/ClimaCore.jl/pull/2114 and https://github.com/CliMA/ClimaComms.jl/pull/103) changed how contexts are adapted to GPUs. This, coupled with
https://github.com/CliMA/ClimaAtmos.jl/pull/3522, prompted me to look again at the `TimeVaryingInputs0D` struct.

This struct is used in two main ways:
- to set site-level forcings/parameters (in ClimaLand)
- to set fields that have constant value on the globe (e.g., CO2)

The previous implementation achieved GPU compatibility by moving everything to the GPU. This is not a great idea when the operation being performed is some variation of linear interpolation in one dimension (time) and for just one data point, as there is nothing really to be parallelized. This choice was made because initially this feature was only used in single-site ClimaLand runs, where all we cared about was to be able to run on GPU. The other day, I realized that I could have GPU compatibility in a much simpler and more robust way by leveraging the `fill!` function. With this commit, 0D interpolation (1D in time, 0D spatially) is changed to be always on the CPU, and the result is written to the Field (or destination in general) using `fill!`.

As a result, the CUDA extension is no longer needed.


charleskawczynski (approved):
Looks good!

PR #149, opened 2025-01-28
Add ITime

ph-kev:
This PR adds ITime which is needed for doing long runs without accumulating floating point errors.

Closes #124


docs/src/timemanager.md
```diff
@@ -1,16 +1,267 @@
...
+time1 = ITime(5; epoch = DateTime(2012, 12, 21))
+time2 = ITime(15; epoch = DateTime(2025, 12, 21))
+try  #hide
+time1 + time2
+catch err; showerror(stderr, err); end  #hide
+```
+
+When dealing with binary operations, it is convenient to think of `ITimes` as
+dimensional quantities (quantities with units). In the previous examples,
+addition returns a new `ITimes`. Division will behave differently
```
Sbozzolo, line 102:
```suggestion
When dealing with binary operations, it is convenient to think of `ITime`s as
dimensional quantities (quantities with units). In the previous examples,
addition returns a new `ITime`s. Division will behave differently 
```

docs/src/timemanager.md
```diff
@@ -1,16 +1,267 @@
...
+
+Sometimes, one need to work with fractions of a `period`. The primary
+application of this is timestepping loops, which are typically divided in stages
+which are a fraction of a timestep.
+
+In these cases, we round to the nearest amount representable as an ITime.
+Furthermore, these cases are constrained to only when an ITime is multipled by a
+float between zero and one. These are the only cases we consider, because the only
+instance in which the problem of handling fraction of a `period` is in the
+timestepping loops.
```
Sbozzolo, line 153:
```suggestion
In these cases, we round to the nearest amount representable as an `ITime`.
Furthermore, these cases are constrained to only when an `ITime` is multipled by a
float between zero and one. These are the only cases we consider, because the only
instance in which the problem of handling fraction of a `period` is in the
timestepping loops.
```

docs/src/timemanager.md
```diff
@@ -1,16 +1,267 @@
...
+0.7 * time # round up
+1.2 * time # error because 1.2 > 1.0
+-0.2 * time # error because -0.2 < 0.0
+```
+
+If multiplication between an ITime and a float is needed, you most likely want
+to cast the ITime into a float. For example, in functions that compute
+tendencies, there are no operations that are of the form `t + a * dt` where `a`
+is a float, so it is safe to cast the ITime into a float as it will not lead to
+a loss of accuracy in time.
```
Sbozzolo, line 169:
```suggestion
If multiplication between an `ITime` and a float is needed, you most likely want
to cast the `ITime` into a float. For example, in functions that compute
tendencies, there are no operations that are of the form `t + a * dt` where `a`
is a float, so it is safe to cast the `ITime` into a float as it will not lead to
a loss of accuracy in time.
```

test/timemanager.jl
```diff
@@ -3,6 +3,66 @@ import Dates
 import CFTime
 using Test
 
+import ClimaTimeSteppers
+import SciMLBase
+
+@testset "TimeStepper" begin
```
Sbozzolo, line 9: skip = pkgversion(ClimaTimeSteppers) >= 0.8

NEWS.md
```diff
@@ -3,6 +3,29 @@ ClimaUtilities.jl Release Notes
 
 main
 ------
+v0.1.21
```
Sbozzolo, line 6:
```suggestion
v0.1.22
```

docs/src/timemanager.md
```diff
@@ -1,16 +1,311 @@
...
+
+## `ITime`
+
+`ITime` is a type to describe times and dates. The "I" in `ITime` stands for
+_integer_: internally, `ITime` uses integers to represent times and dates,
+meaning that operations with `ITime` are exact and do not occur in
+floating-point errors.
+
+`ITime` can be thought a combination of three quantities, a `counter`, a
+`period`, and (optionally) a `epoch`, with the `counter` counting how many
```
Sbozzolo, line 14:
```suggestion
`period`, and (optionally) an `epoch`, with the `counter` counting how many
```

PR #151, opened 2025-02-05
adds xyz regridding interpolation and tet case.

Julians42:
## Purpose 

Adds support for `ClimaCore.Geometry.XYZPoint` to `InterpolationsRegridder` 

## Content

----


ext/InterpolationsRegridderExt.jl
```diff
@@ -77,8 +78,14 @@ function Regridders.InterpolationsRegridder(
...
             (extrapolation_bc = (Intp.Periodic(), Intp.Flat(), Intp.Throw()))
         isnothing(dim_increasing) && (dim_increasing = (true, true, true))
+    elseif eltype(coordinates) <: ClimaCore.Geometry.XYZPoint
+        isnothing(extrapolation_bc) &&
+            (extrapolation_bc = (Intp.Flat(), Intp.Flat(), Intp.Throw()))
+        isnothing(dim_increasing) && (dim_increasing = (true, true, true))
     else
-        error("Only lat-long, lat-long-z spaces are supported")
+        error(
+            "Only lat-long, lat-long-z, and XYZPoint (box) spaces are supported",
```
Sbozzolo, line 87:
```suggestion
            "Only lat-long, lat-long-z, and x-y-z spaces are supported",
```

test/TestTools.jl
```diff
@@ -92,3 +92,51 @@ function make_regional_space(FT; context = ClimaComms.context())
         vertical = vert_center_space,
     )
 end
+
+function make_box_space(FT; context = ClimaComms.context())
```
Sbozzolo, line 96: Could you use `ClimaCore.CommonSpaces`  for this?

(It was introduced later, that's why the other functions in this file don't use it)

test/regridders.jl
```diff
@@ -215,6 +215,50 @@ end
     end
 end
 
+@testset "InterpolationsRegridderXYZPoint" begin
+    helem = (10, 10)
+    Nq = 4
+    zelem = 10
+    FT = Float64 # test for Float64
+
+    x, y, z = collect(0.0:1:10), collect(0.0:1:10), collect(0.0:1:10)
```
Sbozzolo, line 224: Could you make the various dimensions different?

Sbozzolo (commented):
Could you also add an item to the NEWS file?

Sbozzolo:
The MPI tests are failing because of this: https://github.com/CliMA/ClimaCore.jl/issues/2175

Julians42:
@Sbozzolo Thinking we can merge this now as it is working tested on my branch in `ClimaAtmos.jl`. My one remaining question is that if I end up merging the ClimaAtmos time varying forcing code, which requires line 83 to interpolate as `Flat`, `Flat`, `Linear` since we need to interpolate points closer to the ground from ERA5 to the ClimaAtmos.jl grid, we won't be able to test it. Thoughts?

Sbozzolo:
We don't need to test all the variants for `extrapolation_bc`. That's something that is tested elsewhere, so what you have here is fine.

Sbozzolo:
Can you please clean up the commit message (at least removing the typo)?

Sbozzolo (approved):
Awesome! Thank you!

PR #157, opened 2025-02-24
Fix type instability

ph-kev:
This fixes the type instability of `DataHandling.regridded_snapshot` and `Regridders.regrid`.

### DIFF
--- a/ext/InterpolationsRegridderExt.jl
+++ b/ext/InterpolationsRegridderExt.jl
@@ -7,6 +7,7 @@ import ClimaCore.Fields: Adapt
 import ClimaCore.Fields: ClimaComms
 
 import ClimaUtilities.Regridders
+import ClimaUtilities.Utils: unwrap
 
 struct InterpolationsRegridder{
     SPACE <: ClimaCore.Spaces.AbstractSpace,
@@ -119,6 +120,14 @@ Regrid the given data as defined on the given dimensions to the `target_space` i
 This function is allocating.
 """
 function Regridders.regrid(regridder::InterpolationsRegridder, data, dimensions)
+    num_data_dims = ndims(data)
+    num_dims = length(dimensions)
+    num_space_dims = unwrap(regridder.num_space_dims)
+    ((num_space_dims != num_data_dims) || (num_space_dims != num_dims)) &&
+        error(
+            "Number of dimensions of data ($num_data_dims) does not match the dimension of the space ($num_space_dims) or the number of dimensions passed in ($num_dims)",
+        )
+
     FT = ClimaCore.Spaces.undertype(regridder.target_space)
     dimensions_FT = ntuple(
         i ->
--- a/test/data_handling.jl
+++ b/test/data_handling.jl
@@ -144,13 +144,10 @@ ClimaComms.init(context)
         target_space;
         regridder_type = :InterpolationsRegridder,
         file_reader_kwargs = (; preprocess_func = (data) -> 0.0 * data),
-        regridder_kwargs = (;
-            extrapolation_bc = (Intp.Flat(), Intp.Flat(), Intp.Flat())
-        ),
+        regridder_kwargs = (; extrapolation_bc = (Intp.Flat(), Intp.Flat())),
     )
 
-    @test data_handler.regridder.extrapolation_bc ==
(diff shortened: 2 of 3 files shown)


ext/InterpolationsRegridderExt.jl
```diff
@@ -103,16 +118,17 @@ This function is allocating.
 """
 function Regridders.regrid(regridder::InterpolationsRegridder, data, dimensions)
     FT = ClimaCore.Spaces.undertype(regridder.target_space)
-    dimensions_FT = map(dimensions, regridder.dim_increasing) do dim, increasing
-        !increasing ? reverse(FT.(dim)) : FT.(dim)
-    end
+    dimensions_FT = ntuple(
```
Sbozzolo, line 121: Can you also add a check to check that the number of dimensions match between the given data and the space?

ext/InterpolationsRegridderExt.jl
```diff
@@ -27,6 +29,12 @@ struct InterpolationsRegridder{
     """Tuple of booleans signifying if the dimension is monotonically increasing. True for
     dimensions that are monotonically increasing, false for dimensions that are monotonically decreasing."""
     dim_increasing::DT
+
+    """Tuple of integers indicating which dimensions to reverse in data"""
+    decreasing_indices::DI
+
+    "Number of dimensions"
```
charleskawczynski, line 36: What is this the number of dimensions for?

ph-kev: It is the number of dimensions of the space.

src/Utils.jl
```diff
@@ -412,4 +412,11 @@ function sort_by_creation_time(files)
     return sort(files, by = x -> stat(x).ctime)
 end
 
+"""
+    unwrap(x::Val{N})
+
+Unwrap the value in `Val{N}` to get the value at runtime.
```
charleskawczynski, line 418:
```suggestion
Unwrap the value in `Val{N}` to get `N`.
```
It's a compile-time parameter, not really a runtime parameter

ext/InterpolationsRegridderExt.jl
```diff
@@ -27,6 +30,12 @@ struct InterpolationsRegridder{
     """Tuple of booleans signifying if the dimension is monotonically increasing. True for
     dimensions that are monotonically increasing, false for dimensions that are monotonically decreasing."""
     dim_increasing::DT
+
+    """Tuple of integers indicating which dimensions to reverse in data"""
+    decreasing_indices::DI
+
+    "Number of dimensions of the target space"
+    num_space_dims::N
```
charleskawczynski, line 38: Let's eliminate this, and instead write a function that computes this on the fly:

```julia
num_space_dims(regridder::InterpolationsRegridder) =
    num_space_dims(regridder.target_space)
num_space_dims(space::Spaces.AbstractSpace) =
    num_space_dims(eltype(Fields.coordinate_field(space)))
num_space_dims(::Type{S}) where {S <: Geometry.LatLongPoint} = 2
num_space_dims(::Type{S}) where {S <: Geometry.LatLongZPoint} = 3
num_space_dims(::Type{S}) where {S <: Geometry.XYZPoint} = 3
```

It'd be nice if we could do the same for the `decreasing_indices`, but I don't think it's possible since it depends on values of the input data.

charleskawczynski: This should also allow us to eliminate `unwrap` and its unit tests. We could add unit tests for this, of course, instead.

ext/InterpolationsRegridderExt.jl
```diff
@@ -102,17 +120,26 @@ Regrid the given data as defined on the given dimensions to the `target_space` i
...
     FT = ClimaCore.Spaces.undertype(regridder.target_space)
-    dimensions_FT = map(dimensions, regridder.dim_increasing) do dim, increasing
-        !increasing ? reverse(FT.(dim)) : FT.(dim)
-    end
+    dimensions_FT = ntuple(
+        i ->
+            !regridder.dim_increasing[i] ? reverse(FT.(dimensions[i])) :
+            FT.(dimensions[i]),
+        regridder.num_space_dims,
+    )
```
charleskawczynski, line 137: Is there a more elegant way that we can write this?

PR #163, opened 2025-03-21
Add regrid! to InterpolationsRegridder

### LINKED ISSUE #156: Add non-allocating `regrid` in `InterpolationsRegridder`
The only way to regrid is currently `Regridders.regrid`, which does not have an in-place counterpart.

imreddyTeja:
regrid! stores the regridded data in a pre
allocated ClimaCore field. It also uses
interpolate! instead of interpolate. This still
allocates, but much less. It also may
modify the data passed in, but it does not
seem like that happens when using gridded
interpolation. The function also
requires that the base element type of
the passed in dimensions, undertype of the target space, and eltype of the output field all be the same

## Purpose 

Closes #156 

## Content

```
infil> @benchmark Regridders.regrid($reg_horz, $data_lon2D, $dimensions2D)
BenchmarkTools.Trial: 10000 samples with 1 evaluation per sample.
 Range (min … max):  104.167 μs …  10.438 ms  ┊ GC (min … max): 0.00% … 98.75%
 Time  (median):     122.667 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   135.574 μs ± 248.607 μs  ┊ GC (mean ± σ):  9.17% ±  5.12%

                            ▃▇██▆▄▁                              
  ▁▁▁▁▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▂▃▇███████▇▅▄▃▂▂▂▂▂▂▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▂
  104 μs           Histogram: frequency by time          142 μs <

 Memory estimate: 535.33 KiB, allocs estimate: 54.

infil> @benchmark Regridders.regrid!(
                          $in_place_regridded_lon,
                          $reg_horz,
                          $data_lon2D,
                          $dimensions2D,
                      )
BenchmarkTools.Trial: 10000 samples with 1 evaluation per sample.
 Range (min … max):   93.250 μs …   7.032 ms  ┊ GC (min … max): 0.00% … 98.17%
 Time  (median):     102.958 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   111.098 μs ± 203.899 μs  ┊ GC (mean ± σ):  6.82% ±  3.66%

                  ▁▃▅▆▇▇███▇▇▇▆▅▅▄▄▃▃▂▂▁▁▁▁                     ▃
  ▂▃▅▆███████▇▅▅▆▇████████████████████████████▇██▇▇█▇▆▆▆▆▅▆▆▆▆▆ █
  93.2 μs       Histogram: log(frequency) by time        117 μs <

 Memory estimate: 270.70 KiB, allocs estimate: 31.
```

### the `regrid!` that uses `interpolate!` instead of `interpolate`

```

infil> @benchmark Regridders.regrid!(
                   $in_place_regridded_lon,
                   $reg_horz,
                   $data_lon2D,
                   $dimensions2D,
               )
BenchmarkTools.Trial: 10000 samples with 1 evaluation per sample.
 Range (min … max):  86.750 μs … 131.125 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     88.125 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   89.018 μs ±   2.614 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

     ▂█▆▃▂▄                                                     
  ▁▃▆███████▆▃▃▂▂▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▂▃▃▃▃▂▂▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▂
  86.8 μs         Histogram: frequency by time         98.6 μs <

 Memory estimate: 15.41 KiB, allocs estimate: 29.
```

----

### DIFF
--- a/ext/InterpolationsRegridderExt.jl
+++ b/ext/InterpolationsRegridderExt.jl
@@ -132,4 +132,28 @@ function Regridders.regrid(regridder::InterpolationsRegridder, data, dimensions)
     end
 end
 
+"""
+     Regridders.regrid!(output, regridder::InterpolationsRegridder, data::AbstractArray{FT, N}, dimensions::NTuple{N, AbstractVector{FT}})
+
+Regrid the given data as defined on the given dimensions to the `target_space` in `regridder`, and store the result in `output`.
+This method does not automatically reverse dimensions, so the dimensions must be sorted before calling this method.
+"""
+function Regridders.regrid!(
+    output,
+    regridder::InterpolationsRegridder,
+    data::AbstractArray{FT, N},
+    dimensions::NTuple{N, AbstractVector{FT}},
+) where {FT, N}
+    all(regridder.dim_increasing) || error(
+        "Dimensions must be monotonically increasing to use regrid!. Sort the dimensions first, or use regrid.",
+    )
+    itp = Intp.extrapolate(
+        Intp.interpolate(dimensions, data, Intp.Gridded(Intp.Linear())),
+        regridder.extrapolation_bc,
+    )
+    gpuitp = Adapt.adapt(ClimaComms.array_type(regridder.target_space), itp)
+    output .= splat(gpuitp).(totuple.(regridder.coordinates))
+    return nothing
+end
+
 end
--- a/src/Regridders.jl
+++ b/src/Regridders.jl
@@ -23,6 +23,8 @@ function InterpolationsRegridder end
 
 function regrid end
 
+function regrid! end
+
 """
     default_regridder_type()
 
--- a/test/regridders.jl
+++ b/test/regridders.jl
@@ -129,6 +129,13 @@ end
     @test regridded_lat_reversed == regridded_lat
     @test regridded_lon_reversed == regridded_lon
     @test regridded_z_reversed == regridded_z
+
+    @test_throws "Dimensions must be monotonically increasing to use regrid!. Sort the dimensions first, or use regrid." Regridders.regrid!(
+        zeros(axes(regridded_z_reversed)),
+        reg_hv_reversed,
+        data_lat3D,
+        dimensions3D_reversed,
+    )
 end
 
 @testset "InterpolationsRegridder" begin
@@ -177,6 +184,25 @@ end
         regridded_lat = Regridders.regrid(reg_horz, data_lat2D, dimensions2D)
         regridded_lon = Regridders.regrid(reg_horz, data_lon2D, dimensions2D)
 
+        # now repeat the regrid with in place method, and check that the result is the same
+        in_place_regridded_lat = zeros(axes(regridded_lat))
+        in_place_regridded_lon = zeros(axes(regridded_lon))
+        Regridders.regrid!(
+            in_place_regridded_lat,
+            reg_horz,
+            FT.(data_lat2D),
+            map(x -> FT.(x), dimensions2D),
+        )
+        Regridders.regrid!(
+            in_place_regridded_lon,
+            reg_horz,
+            FT.(data_lon2D),
+            map(x -> FT.(x), dimensions2D),
+        )
+
+        @test regridded_lat == in_place_regridded_lat
+        @test regridded_lon == in_place_regridded_lon
+
         coordinates = ClimaCore.Fields.coordinate_field(horzspace)
 
         # Compute max err
@@ -201,6 +227,32 @@ end
         regridded_lat = Regridders.regrid(reg_hv, data_lat3D, dimensions3D)
         regridded_lon = Regridders.regrid(reg_hv, data_lon3D, dimensions3D)
         regridded_z = Regridders.regrid(reg_hv, data_z3D, dimensions3D)
+
+        in_place_regridded_lat = zeros(axes(regridded_lat))
+        in_place_regridded_lon = zeros(axes(regridded_lon))
+        in_place_regridded_z = zeros(axes(regridded_z))
+        Regridders.regrid!(
+            in_place_regridded_lat,
+            reg_hv,
+            FT.(data_lat3D),
+            map(x -> FT.(x), dimensions3D),
+        )
+        Regridders.regrid!(
+            in_place_regridded_lon,
+            reg_hv,
+            FT.(data_lon3D),
+            map(x -> FT.(x), dimensions3D),
+        )
+        Regridders.regrid!(
+            in_place_regridded_z,
+            reg_hv,
+            FT.(data_z3D),
+            map(x -> FT.(x), dimensions3D),
+        )
+
+        @test regridded_lat == in_place_regridded_lat
+        @test regridded_lon == in_place_regridded_lon
+        @test regridded_z == in_place_regridded_z
 
         coordinates = ClimaCore.Fields.coordinate_field(hv_center_space)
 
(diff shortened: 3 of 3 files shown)


ext/InterpolationsRegridderExt.jl
```diff
@@ -132,4 +132,33 @@ function Regridders.regrid(regridder::InterpolationsRegridder, data, dimensions)
...
+    @assert eltype(output) == FT
+    @assert eltype(data) == FT
+    all(regridder.dim_increasing) || error(
+        "Dimensions must be sorted to use regrid!. Sort the dimensions first, or use regrid.",
+    )
+    # interpolate! still allocates some, and might "destroy" data while using the array to store
+    # coefficients... but it seems like the data is equal to the coefficients
+    # for a gridded interpolation, so data is not modified
+    itp = Intp.extrapolate(
+        Intp.interpolate!(dimensions, data, Intp.Gridded(Intp.Linear())),
```
imreddyTeja, line 157: I think this does not modify `data`, but I need to confirm this.

See [here](https://github.com/JuliaMath/Interpolations.jl/blob/75786416f29de9e332551880e789bd7ccf7d947a/src/gridded/gridded.jl#L178-L179)

ext/InterpolationsRegridderExt.jl
```diff
@@ -132,4 +132,33 @@ function Regridders.regrid(regridder::InterpolationsRegridder, data, dimensions)
...
+"""
+function Regridders.regrid!(
+    output,
+    regridder::InterpolationsRegridder,
+    data::AbstractArray{FT, N},
+    dimensions::NTuple{N, AbstractVector{FT}},
+) where {FT, N}
+    @assert ClimaCore.Spaces.undertype(regridder.target_space) == FT
+    @assert eltype(output) == FT
+    @assert eltype(data) == FT
```
imreddyTeja, line 149: Is it better to have these asserts, or let type conversions/errors occur?

ext/InterpolationsRegridderExt.jl
```diff
@@ -132,4 +132,33 @@ function Regridders.regrid(regridder::InterpolationsRegridder, data, dimensions)
     end
 end
 
+"""
+     Regridders.regrid!(output, regridder::InterpolationsRegridder, data::AbstractArray{FT, N}, dimensions::NTuple{N, AbstractVector{FT}})
+
+Regrid the given data as defined on the given dimensions to the `target_space` in `regridder`, and store the result in `output`.
+This method does not automatically reverse dimensions, so the dimensions must be sorted before calling this method.
```
imreddyTeja, line 139: This could be supported, but would either need to modify `dimensions` and `regridder`, or allocate a new dimensions tuple of vectors.

ext/InterpolationsRegridderExt.jl
```diff
@@ -132,4 +132,38 @@ function Regridders.regrid(regridder::InterpolationsRegridder, data, dimensions)
...
+    # interpolate! still allocates some, and might "destroy" data while using the array to store
+    # coefficients... but it seems like the data is equal to the coefficients
+    # for a gridded interpolation, so data is not modified
+    itp = Intp.extrapolate(
+        Intp.interpolate!(dimensions, data, Intp.Gridded(Intp.Linear())),
+        regridder.extrapolation_bc,
+    )
+    gpuitp = Adapt.adapt(ClimaComms.array_type(regridder.target_space), itp)
+    if pkgversion(ClimaCore) < v"0.14.27"
+        @warn "regrid! allocates when using ClimaCore < 0.14.27"
```
ph-kev, line 162: I think it would be annoying to see this warning every time that you call this function. I would move it to the description of the function and add a warning admonition.

ext/InterpolationsRegridderExt.jl
```diff
@@ -132,4 +132,38 @@ function Regridders.regrid(regridder::InterpolationsRegridder, data, dimensions)
...
+    output,
+    regridder::InterpolationsRegridder,
+    data::AbstractArray{FT, N},
+    dimensions::NTuple{N, AbstractVector{FT}},
+) where {FT, N}
+    @assert ClimaCore.Spaces.undertype(regridder.target_space) == FT
+    @assert eltype(output) == FT
+    @assert eltype(data) == FT
+    all(regridder.dim_increasing) || error(
+        "Dimensions must be sorted to use regrid!. Sort the dimensions first, or use regrid.",
```
ph-kev, line 151: For this error message, I think it would be more accurate to say that the values need to be in increasing order. For example, the values can be in decreasing order and still be sorted.

ext/InterpolationsRegridderExt.jl
ph-kev: I am confused on why we need to enforce this? What happen if the types are not the same? Wouldn't the type promotes?

ext/InterpolationsRegridderExt.jl
ph-kev: If it's not too much work, can't you write a test that check if data is the same before and after calling `regrid!`.

ext/InterpolationsRegridderExt.jl
```diff
@@ -132,4 +132,30 @@ function Regridders.regrid(regridder::InterpolationsRegridder, data, dimensions)
...
+    )
+    # interpolate! still allocates some, and might "destroy" data while using the array to store
+    # coefficients... but it seems like the data is equal to the coefficients
+    # for a gridded interpolation, so data is not modified
+    itp = Intp.extrapolate(
+        Intp.interpolate!(dimensions, data, Intp.Gridded(Intp.Linear())),
+        regridder.extrapolation_bc,
+    )
+    gpuitp = Adapt.adapt(ClimaComms.array_type(regridder.target_space), itp)
+    output .= splat(gpuitp).(totuple.(regridder.coordinates))
```
Sbozzolo, line 158:
```suggestion
    output .= splat(gpuitp).(totuple.(regridder.coordinates))
    return nothing
```

Cool, I didn't know about `splat`!

ext/InterpolationsRegridderExt.jl
```diff
@@ -132,4 +132,30 @@ function Regridders.regrid(regridder::InterpolationsRegridder, data, dimensions)
...
+    dimensions::NTuple{N, AbstractVector{FT}},
+) where {FT, N}
+    all(regridder.dim_increasing) || error(
+        "Dimensions must be monotonically increasing to use regrid!. Sort the dimensions first, or use regrid.",
+    )
+    # interpolate! still allocates some, and might "destroy" data while using the array to store
+    # coefficients... but it seems like the data is equal to the coefficients
+    # for a gridded interpolation, so data is not modified
+    itp = Intp.extrapolate(
+        Intp.interpolate!(dimensions, data, Intp.Gridded(Intp.Linear())),
```
Sbozzolo, line 154: Why do you need `interpolate!`?

imreddyTeja: It is not needed at all, and is now changed to `interpolate`. I think with this specific use case, `interpolate` could be replaced by `interpolate!` with no real cost. `interpolate!` has significantly less allocations than `interpolate`, but may overwrite the input data (it does not in this case). 


I originally used `interpolate!` to make the overall `regrid` function have less allocations. Similarly, I also made `regrid` not automatically sort the dimensions and covert to the correct type to avoid allocating new data and dimensions arrays. Should that also be changed?

imreddyTeja: Difference between using `interpolate` and `interpolate!`:

```

infil> @benchmark Regridders.regrid!(
                   $in_place_regridded_lon,
                   $reg_horz,
                   $data_lon2D,
                   $dimensions2D,
               )
BenchmarkTools.Trial: 10000 samples with 1 evaluation per sample.
 Range (min … max):  86.750 μs … 131.125 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     88.125 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   89.018 μs ±   2.614 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

     ▂█▆▃▂▄                                                     
  ▁▃▆███████▆▃▃▂▂▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▂▃▃▃▃▂▂▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▂
  86.8 μs         Histogram: frequency by time         98.6 μs <

 Memory estimate: 15.41 KiB, allocs estimate: 29.
```

```
@benchmark Regridders.regrid!(
                          $in_place_regridded_lon,
                          $reg_horz,
                          $data_lon2D,
                          $dimensions2D,
                      )
BenchmarkTools.Trial: 10000 samples with 1 evaluation per sample.
 Range (min … max):   93.250 μs …   7.032 ms  ┊ GC (min … max): 0.00% … 98.17%
 Time  (median):     102.958 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   111.098 μs ± 203.899 μs  ┊ GC (mean ± σ):  6.82% ±  3.66%

                  ▁▃▅▆▇▇███▇▇▇▆▅▅▄▄▃▃▂▂▁▁▁▁                     ▃
  ▂▃▅▆███████▇▅▅▆▇████████████████████████████▇██▇▇█▇▆▆▆▆▅▆▆▆▆▆ █
  93.2 μs       Histogram: log(frequency) by time        117 μs <

 Memory estimate: 270.70 KiB, allocs estimate: 31.
```

Sbozzolo (approved):

PR #165, opened 2025-05-13
Add deferopen = false when constructing multifile dataset

ph-kev:
This PR adds `deferopen = false` when constructing a multifile dataset. According to this issue (https://github.com/JuliaGeo/NCDatasets.jl/issues/277), files are not kept open because of the limitation of 1024 open files on Linux. However, for our use case, this will not matter. 

See this [buildkite](https://buildkite.com/clima/climaland-long-runs/builds/3971#0196b1a7-1aae-4a0c-9df4-6ce556fdc33b) for a ClimaLand run for a single year with ~40 years of ERA5 forcing data without this change.
See this [buildkite](https://buildkite.com/clima/climaland-long-runs/builds/3987#0196cad7-4762-452d-b592-899dc75cc2f0) for a ClimaLand run for a single year with ~40 years of ERa5 forcing data with this change.
Also, see this [buildkite](https://buildkite.com/clima/climaland-long-runs/builds/3977#0196b251-76f1-4d94-95db-2e9aa179c48c) for a ClimaLand run for a single year with a single year of ERA5 forcing data.

The wall time per step for the three runs are

1. 35 milliseconds, 834 microseconds (forty years of ERA5 forcing data)
2. 25 milliseconds, 126 microseconds (forty years of ERA5 forcing data with this branch)
3. 22 milliseconds, 240 microseconds (single year of ERA5 forcing data)

### DIFF
--- a/ext/DataHandlingExt.jl
+++ b/ext/DataHandlingExt.jl
@@ -581,7 +581,6 @@ function DataHandling.regridded_snapshot(
     end
 
     regridder_type = nameof(typeof(data_handler.regridder))
-    regrid_args = ()
 
     # Check if the regridded field at this date is already in the cache
     return get!(data_handler._cached_regridded_fields, date) do
--- a/ext/NCFileReaderExt.jl
+++ b/ext/NCFileReaderExt.jl
@@ -107,7 +107,13 @@ function FileReaders.NCFileReader(
             is_time = x -> x == "time" || x == "date" || x == "t"
             time_dims = filter(is_time, NCDatasets.dimnames(first_dataset))
             if !isempty(time_dims)
-                aggtime_kwarg = (:aggdim => first(time_dims),)
+                # When loading multifile dataset using NCDatasets.jl, the NetCDF files are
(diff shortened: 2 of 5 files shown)


.github/workflows/JuliaFormatter.yml
```diff
@@ -40,7 +40,7 @@ jobs:
     - name: Apply JuliaFormatter
       if: steps.filter.outputs.julia_file_change == 'true'
       run: |
-        julia -e 'using Pkg; Pkg.add("JuliaFormatter"); using JuliaFormatter; format(".")'
+        julia -e 'using Pkg; Pkg.add(name = "JuliaFormatter", version = "1.0.62"); using JuliaFormatter; format(".")'
```
ph-kev, line 44: I wasn't too sure if we want to keep using 1.0.62 or use the latest version of JuliaFormatter which introduces minor formatting differences compared to before.

Sbozzolo: Let's use the lastest formatter

Sbozzolo (approved):

Issue #166, opened 2025-06-20
Error within NCFileReader error message


imreddyTeja:
If attempting to open a set of files with non-matching dimensions, an error is thrown with
`error("$file_path does not contain information about dimensions")`. This throws 
`ERROR: LoadError: UndefVarError: `file_path` not defined in `ClimaUtilitiesNCDatasetsExt.NCFileReaderExt`
instead.

imreddyTeja:
This same issue is found in other error messages

PR #167, opened 2025-06-23
Fix NCFileReader error messages

imreddyTeja:
## Purpose 
Cloes #166 

PR #127 added support for reading from multiple netcdf files. In the PR, the variable
`file_name` is changed to `file_names`. Some of the error messages do not reflect this
change, and will throw a `UndefVarError` instead of the intended error.

## Content
- add a test for behavior when reading a variable with dimensions not found in the dataset
- fix error messages with `file_name`, and update to reflect multiple file reading

----


test/file_readers.jl
```diff
@@ -117,3 +117,16 @@ end
         @test read_dates_func(nc) == DateTime.(string.(times), "yyyymmdd")
     end
 end
+
+@testset "read_missing_dims" begin
+    data_dir = mktempdir()
+    @test_throws contains(
+        "test_dims.nc\"] does not contain information about dimensions",
+    ) NCDataset(joinpath(data_dir, "test_dims.nc"), "c") do nc
+        defVar(nc, "some_var", [1, 2], ("invalid_dim",))
```
ph-kev, line 126: Would it be possible to use the already existing file in `test_data` directory or make a new file and place it there?

ph-kev (approved):

PR #176, opened 2025-07-24
Allow regridding 2D data onto 2D spaces with LatLongZ coordinates

### LINKED ISSUE #174: `DataHandler` putting 2D data onto a 3D space fails with an unhelpful message
When trying to read 2D NCDataset onto any space that uses LatLongZPoint, an uninformative error is raised. 

This is troublesome because in ClimaLand, we construct a PointSpace that uses LatLongZPoint, and we would like to interpolate 2D surface data onto the lat/long at that point (and ignore the z dimension).

Ideally, I would want to check the dimension names of the input data against the dimensions of the space's coordinate field. If they're both lat/long, we can ignore the z dimension and regrid in the horizontal only.

At the very least, if we don't want to support this case we should raise an error explaining the dimension mismatch between input data and the target space.

### Error message
```
ERROR: MethodError: no method matching (::Interpolations.Extrapolation{…})(::Float64, ::Float64, ::Float64)
Use square brackets [] for indexing an Array.
The object of type `Interpolations.Extrapolation{Float64, 2, Interpolations.GriddedInterpolation{Float64, 2, Matrix{Float64}, Interpolations.Gridded{Interpolations.Linear{Interpolations.Throw{Interpolations.OnGrid}}}, Tuple{Vector{Float64}, Vector{Float64}}}, Interpolations.Gridded{Interpolations.Linear{Interpolations.Throw{Interpolations.OnGrid}}}, Tuple{Interpolations.Periodic{Nothing}, Interpolations.Flat{Nothing}, Interpolations.Throw{Nothing}}}` exists, but no method is defined for this combination of argument types when trying to treat it as a callable object.
```

## Reproducer
Run from ClimaUtilities.jl/test/
```julia
using Artifacts
import ClimaUtilities
import ClimaUtilities.DataHandling
import ClimaUtilities.FileReaders
import ClimaCore
import ClimaComms
@static pkgversion(ClimaComms) >= v"0.6" && ClimaComms.@import_required_backends
import Interpolations as Intp # uses InterpolationsRegridder
using NCDatasets
include("TestTools.jl")

context = ClimaComms.context()
ClimaComms.init(context)
spaces = make_spherical_space(Float64; context)
hv_center_space = spaces.hybrid # ExtrudedFiniteDifferenceSpace with LatLongZ coordinates

file_path_2d = joinpath(
        artifact"era5_static_example",
        "era5_t2m_sp_u10n_20210101_static.nc",
    )
varname_2d = ["sp"]
data_handler_2d = DataHandling.DataHandler(file_path_2d, varname_2d, hv_center_space)
DataHandling.regridded_snapshot(data_handler_2d) # THIS FAILS

file_path_3d = joinpath(@__DIR__, "test_data", "sorted_dims.nc")
varname_3d = ["test_var"]
data_handler_3d = DataHandling.DataHandler(file_path_3d, varname_3d, hv_center_space)
DataHandling.regridded_snapshot(data_handler_3d) # THIS WORKS
```

juliasloan25:
## Purpose 
Previously, this would error because the interpolator would be constructed with dimensions corresponding to the 2D input data, but then try to regrid onto the 3D coordinates. The solution in this PR is to handle that case by displaying a warning and dropping the z dimension while interpolating.

Closes #174

## Content

----

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,12 @@ ClimaUtilities.jl Release Notes
 main
 ------
 
+#### Allow regridding 2D data onto 2D spaces with LatLongZ coordinates. PR[#176](https://github.com/CliMA/ClimaUtilities.jl/pull/176)
+
+Some 2D spaces use `LatLongPoint` coordinates and others use `LatLongZPoint`,
+depending on how they were constructed. Both cases are now supported, so we can
+read in 2D data onto a 2D space with either coordinate type.
+
 #### Interpolation method for InterpolationsRegridder
 
 `InterpolationsRegridder` now accepts the new keyword argument
@@ -29,9 +35,9 @@ v0.1.23
 
 #### Support for Box Interpolations Regridder. PR[#151](https://github.com/CliMA/ClimaUtilities.jl/pull/151)
 
-`Regridders.IterpolationsRegridder` now supports regridding on 
-`ClimaCore.Geometry.XYZPoint` objects which allows for interpolation 
-onto boxes and single column simulations. 
+`Regridders.IterpolationsRegridder` now supports regridding on
+`ClimaCore.Geometry.XYZPoint` objects which allows for interpolation
+onto boxes and single column simulations.
 
 v0.1.22
 ------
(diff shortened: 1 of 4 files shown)


test/data_handling.jl
```diff
@@ -541,3 +541,27 @@ end
 
     end
 end
+
+@testset "DataHandler, 2D space with LatLongZ coordinates" begin
```
juliasloan25, line 545: delete this

imreddyTeja (approved):

PR #177, opened 2025-07-24
fix tests for longitude regridding

### LINKED ISSUE #175: InterpolationsRegridder has large errors in test
The InterpolationsRegridder tests regrid coordinates (lat, lon, z) onto a ClimaCore space and check that the difference when compared to the space's coordinates are small. However in the check we forget to take an absolute value, so there have been large errors that were missed previously (up to 180 or 360). See an example [here](https://github.com/CliMA/ClimaUtilities.jl/blob/67a97c88f5ed884a6cc0bdb3604405b63b4d0953/test/regridders.jl#L209-L213).

Notably, large errors only appear in the regridded longitude. Even after changing the `extrapolation_bc` for all dimensions to be `Flat`, longitude is still the only one with large errors. Maybe because it has all positive values, where latitude is in [-90, 90]?

juliasloan25:
## Purpose 
The InterpolationsRegridder tests have been checking that the maximum difference between regridded and reference values is below a threshold, but without taking the absolute value of the difference. This led tests to pass when they should have failed.

Closes #175

## Content
- [x] take absolute value of difference in tests
- [x] use longitude on [-180, 180] instead of [0, 360] for comparison with ClimaCore coordinates
- [x] check that values at lon=180 are the same as lon=-180 with periodic boundary conditions

### DIFF
--- a/test/regridders.jl
+++ b/test/regridders.jl
@@ -31,7 +31,7 @@ end
 
 @testset "InterpolationsRegridder incorrect dimensions" begin
     lon, lat, z =
-        collect(0.0:1:360), collect(-90.0:1:90), collect(0.0:1.0:100.0)
+        collect(-180.0:1:180.0), collect(-90.0:1:90), collect(0.0:1.0:100.0)
     dimensions3D = (lon, lat, z)
     dimensions3D_reversed = (lon, reverse(lat), reverse(z))
     dimensions2D = (lon, lat)
@@ -141,7 +141,7 @@ end
 @testset "InterpolationsRegridder" begin
 
     lon, lat, z =
-        collect(0.0:1:360), collect(-90.0:1:90), collect(0.0:1.0:100.0)
+        collect(-180.0:1:180.0), collect(-90.0:1:90), collect(0.0:1.0:100.0)
     dimensions2D = (lon, lat)
     dimensions3D = (lon, lat, z)
     size2D = (361, 181)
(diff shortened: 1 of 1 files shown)


nefrathenrici (commented):
This looks good to me, but I am not familiar with this code. The duplicated testing logic could be put into a helper function if we want to extend these tests in the future.

test/regridders.jl
```diff
@@ -206,11 +206,26 @@ end
...
-        err_lat = coordinates.lat .- regridded_lat
-        err_lon = coordinates.long .- regridded_lon
-
+        err_lat = abs.(coordinates.lat .- regridded_lat)
         @test maximum(err_lat) < 1e-5
-        @test maximum(err_lon) < 1e-5
+
+        # Since lon uses periodic BCs, error is large at 180 degrees.
+        # We check that the error is small at other indices, and that the values at -180 and 180 agree.
+        # Note that for data that is actually periodic, the error at 180 degrees will also be small.
```
ph-kev, line 214: It is usually not the case that the error will be small, because the longitudes in other NetCDF files are usually from -179.5 to 179.5 or 0.5 to 359.5 or some variant of that. This is because they represent the center of the cells. As a result, any interpolation at the boundaries is wrong.

ph-kev (approved):

test/regridders.jl
juliasloan25: Shouldn't the periodic extrapolation boundary condition take care of that? E.g. if the data at -179.5 and 179.5 is similar in values, we'll use periodic extrapolation to fill the value at 180 with something reasonable

ph-kev: It's slightly inaccurate. Since we are using a gridded interpolation, if we want data at 180, it gives us the data at -179. This is because 179.5 + 0.5 = 180.0 and -179.5 + 0.5 = -179.

juliasloan25: Okay got it. I'll update the comment here

PR #178, opened 2025-07-25
ITime housekeeping

imreddyTeja:
Add tests for TimeVaryingInput0D for 0D TVI
with times of floats, ITime with no epoch, and ITime with epoch.

Tests 0d TVI of floats can evaluate at ITime inputs, floating point inputs, and tests that an error is thrown with a datetime input.

Test that error is thrown when evaluating a TVI of ITimes with no epoch at a DateTime.

Test that TVI of ITime with epoch can be evaluated at ITimes, floats, or DateTimes.

Test that TVI of ITime with no epoch can be evaluated at ITimes or floats.

Make all the above tests pass by adding conversions to the `evaluate!` methods.

Add `Float64(::ITime)` and `Float32(::Time)`

Tried changing ITime and floating point  multiplication behavior to attempt to use a smaller period if the result of the multiplication is less than 1, but that is a behavior change, so I reverted it

## Purpose 

----

### DIFF
--- a/ext/TimeVaryingInputs0DExt.jl
+++ b/ext/TimeVaryingInputs0DExt.jl
@@ -20,6 +20,7 @@ import ClimaUtilities.TimeVaryingInputs: extrapolation_bc
 import ClimaUtilities.TimeVaryingInputs
 
 import ClimaUtilities.TimeManager: ITime, date
+using Dates
 
 """
     InterpolatingTimeVaryingInput0D
@@ -27,9 +28,9 @@ import ClimaUtilities.TimeManager: ITime, date
 The constructor for InterpolatingTimeVaryingInput0D is not supposed to be used directly, unless you
 know what you are doing.
 
-`times` and `vales` may have different float types, but they must be the same length, and we
+`times` and `values` may have different types, but they must be the same length, and we
 assume that they have been sorted to be monotonically increasing in time, without repeated
-values for the same timestamp.
+values for the same timestamp. `times` can have elements of type `ITime` or floats
 """
 struct InterpolatingTimeVaryingInput0D{
     AA1 <: AbstractArray,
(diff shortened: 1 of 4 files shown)


test/itime.jl
```diff
@@ -148,6 +148,8 @@ using Test, Dates
     @testset "Float Conversion and Broadcasting" begin
         t1 = ITime(10, period = Dates.Millisecond(100))
         @test float(t1) == 1.0
+        @test Float64(t1) == Float64(1)
+        @test Float32(t1) == Float32(1)
```
ph-kev, line 152: This doesn't check the types (e.g., `Float64(1) == Float32(1)` is true).

ext/TimeVaryingInputs0DExt.jl
```diff
@@ -227,4 +241,53 @@ function TimeVaryingInputs.evaluate!(
...
+
+function TimeVaryingInputs.evaluate!(
+    destination,
+    itp::InterpolatingTimeVaryingInput0D{<:AbstractArray{<:ITime}},
+    time::Number,
+    args...;
+    kwargs...,
+)
+    converted_time = promote(ITime(time), itp.range[1])[1]
+    return TimeVaryingInputs.evaluate!(
```
ph-kev, line 273: This is just style, but I think
```suggestion
    converted_time = first(promote(ITime(time), itp.range[1]))
    return TimeVaryingInputs.evaluate!(
```
is a little cleaner.

ext/TimeVaryingInputs0DExt.jl
```diff
@@ -227,4 +241,53 @@ function TimeVaryingInputs.evaluate!(
...
+    destination,
+    itp::InterpolatingTimeVaryingInput0D{<:AbstractArray{<:ITime}},
+    date::DateTime,
+    args...;
+    kwargs...,
+)
+    t0_date = DateTime(itp.range[1])
+    diff_ms = date - t0_date
+    converted_time =
+        ITime(diff_ms.value; period = typeof(diff_ms)(1), epoch = t0_date)
```
ph-kev, line 254: I am wondering if this is correct? What is the type of `itp.range[1]`?

ph-kev: I think it is fine to loosen the type restriction, because if the date isn't a `DateTime`, then it would fail anyway when you tried to subtract.

ph-kev (approved):

PR #187, opened 2025-08-22
Change TVI0d constructor to conditionally promote

### LINKED ISSUE #186: Using TVI0D with PeriodCalendar is very slow
In our fluxnet simulations, we wish to construct the time varying input from observations. These observations are half hourly, and can last many years. For example, even one year of data consists of 4e5 points. It is convenient to use PeriodicBC (for spinup, etc).

In this case, this line becomes very slow
https://github.com/CliMA/ClimaUtilities.jl/blob/37bb7ab6ada77b749bacb245c9c2719479227cf1/ext/TimeVaryingInputs0DExt.jl#L112C42-L112C72

For example:
```
times = [ClimaUtilities.TimeManager.ITime(x; epoch =  DateTime(2008)) for x in 0:1:100];
@time [float.(promote(times...))...];
  0.256722 seconds (937.28 k allocations: 47.744 MiB, 3.62% gc time, 99.76% compilation time)
times = [ClimaUtilities.TimeManager.ITime(x; epoch =  DateTime(2008)) for x in 0:1:1000];
@time [float.(promote(times...))...];
  1.645448 seconds (3.61 M allocations: 234.321 MiB, 4.09% gc time, 98.11% compilation time)
times = [ClimaUtilities.TimeManager.ITime(x; epoch =  DateTime(2008)) for x in 0:1:10000];
@time [float.(promote(times...))...];
 33.065048 seconds (35.07 M allocations: 8.290 GiB, 3.95% gc time, 90.77% compilation time)
```

Simply converting to a float is much faster:
```
julia> f(x) = ClimaUtilities.float(x)
f (generic function with 1 method)
julia> @time f.(times);
  0.075337 seconds (208.04 k allocations: 10.002 MiB, 99.50% compilation time: 42% of which was recompilation)
julia> length(times)
10001
```

I think the issue must be `promote` (which has to search a Tuple and do a bunch of conversions?)

imreddyTeja:
Promoting a large number of ITimes is expensive, so this modifies the behavior to only do so if needed.

Closes #186

----

### DIFF
--- a/ext/TimeVaryingInputs0DExt.jl
+++ b/ext/TimeVaryingInputs0DExt.jl
@@ -105,12 +105,25 @@ function TimeVaryingInputs.TimeVaryingInput(
             "LinearPeriodFillingInterpolation is not supported when the input data is 1D",
         )
     end
-
+    if eltype(times) <: ITime
+        if !all(
+            t ->
+                t.period == first(times).period &&
+                t.epoch == first(times).epoch,
+            times,
+        )
+            # promote if times do not all have same epoch and period to avoid promoting
+            # during the simulation
+            times = [promote(times...)...]
+        elseif !(eltype(times) <: ITime{<:Any, <:Any, Nothing})
+            all(d -> d.epoch == first(times).epoch, times) || error(
+                "TimeVaryingInputs cannot be used when the data is defined at `ITime`(s) with differing epochs",
+            )
+        end
+    end
     if extrapolation_bc(method) isa PeriodicCalendar
         if extrapolation_bc(method) isa PeriodicCalendar{Nothing}
-            if !isequispaced(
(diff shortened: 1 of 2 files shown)


ext/TimeVaryingInputs0DExt.jl
```diff
@@ -105,12 +105,23 @@ function TimeVaryingInputs.TimeVaryingInput(
             "LinearPeriodFillingInterpolation is not supported when the input data is 1D",
         )
     end
-
+    if eltype(times) <: ITime
+        if !all(
+            t -> t.period == times[1].period && t.epoch == times[1].epoch,
+            times,
```
ph-kev, line 111:
```suggestion
            t -> t.period == first(times).period && t.epoch == first(times).epoch,
            times,
```

ph-kev (approved):

Issue #181, opened 2025-08-01
Gridded vs on cell boundary condition


ph-kev:
For the regridder, we use gridded interpolation:

https://github.com/CliMA/ClimaUtilities.jl/blob/36595eac5d59e5f4285da1739167eccad061b840/ext/InterpolationsRegridderExt.jl#L143-L150

This can leads to a correctness issue when we are interpolating along the longitudes. For example, when interpolating from 0.5 to 359.5 degrees to -180.0 to 180.0 degrees, any values less than 0.5 is not interpolated correctly. For example, if you want -1 degrees, then it should be 359.0 degrees. However, since a gridded interpolation is used, the actual value is 358.0 degrees because 359.5 - 1.5 = 358.0 and 0.5 - 1.5 = -1. 

One solution to this is to see if it is possible to add your own boundary condition to Interpolations.jl that fix this behavior for the longitude coordinates. Another solution is to add this to ClimaInterpolations.jl and use that instead of Interpolations.jl.

PR #190, opened 2025-10-03
Use oncell periodic condition when possible

ph-kev:
This PR uses `OnCell` periodic boundary condition whenever possible. This is done by checking if all the dimensions are equispaced and if they are, ~~use `BSpline`~~ add virtual longitude points by repeating the first longitude slice.

~~All dimensions must be equispaced, because `BSpline` in Interpolations.jl required that. As a result, if the interpolation is from data with time, longitude, and latitude as the dimensions, then a `BSpline` cannot be used. As a result, the interpolation is not correct for the longitude dimension.~~

The proposed solution is to use a heuristic to automatically ~~choose oncell~~ add virtual longitude points if one of the dimensions span 360 degrees by seeing if `last(dim) - first(dim) + dim[2] - dim[1] ` is 360 degrees and the dimension is equispaced.

This is a stopgap solution to #181.

TODO

- [X] Add tests

### DIFF
--- a/ext/InterpolationsRegridderExt.jl
+++ b/ext/InterpolationsRegridderExt.jl
@@ -50,6 +50,7 @@ totuple(pt::ClimaCore.Geometry.ZPoint) = pt.z
     InterpolationsRegridder(target_space::ClimaCore.Spaces.AbstractSpace
                             [; extrapolation_bc::Tuple,
                                dim_increasing::Union{Nothing, Tuple},
+                               dim_names::Union{Nothing, Tuple}
                                interpolation_method = Interpolations.Linear()])
 
 An online regridder that uses Interpolations.jl


ext/InterpolationsRegridderExt.jl
```diff
@@ -212,6 +266,18 @@ function Regridders.regrid(regridder::InterpolationsRegridder, data, dimensions)
     end
 end
 
+"""
+    _append_first_slice(data, dim_idx::Int)
+
+Append the first slice along the `dim_idx`th dimension to the end of the
+`dim_idx`th dimension of `data`.
+"""
+function _append_first_slice(data, dim_idx::Int)
```
imreddyTeja: Should this be `append_first_slice!`?

imreddyTeja (approved):

PR #195, opened 2025-10-17
update docs for PeriodicCalendar

### LINKED ISSUE #194: Document `PeriodicCalendar` more clearly
Right now `PeriodicCalendar` is implemented as an extrapolation BC, which suggests that it works by repeating data only outside the bounds of when the data exists. In reality, when provided a period it repeats that period during the entire simulation, even if data is defined during those times. This should be stated very explicitly in the docs to avoid user confusion.

juliasloan25:
## Purpose 
Clearly document the case where using PeriodicCalendar reuses one period of data, even when other data is available (i.e. not acting as a true "extrapolation boundary condition".

Closes #194

### DIFF
--- a/docs/src/inputs.md
+++ b/docs/src/inputs.md
@@ -137,6 +137,16 @@ the duration of the period that has to be repeated. The `repeat_date` defines
 what date range needs to be repeated. For example, if `period = Dates.Month(1)`
 and `repeat_date = Dates.Date(1993, 11)`, November 1993 will be repeated.
 
+!!! warn
+
+    While `PeriodicCalendar` is described as an extrapolation boundary condition,
+    it may also change the data used during the time period where data is available.
+    For example, take the case where data from 2000 to 2020 is available, and we have
+    `period = Dates.Year(1)` and `repeat_date = Dates.Date(2000)`. The data from 2000
+    will be repeated for *every year of the simulation* - outside of the 2000-2020
+    range where data is available, but also for every year within that range.
+    In this sense, `PeriodicCalendar` is not a true boundary condition only.
+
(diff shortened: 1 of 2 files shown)


docs/src/inputs.md
```diff
@@ -137,6 +137,16 @@ the duration of the period that has to be repeated. The `repeat_date` defines
 what date range needs to be repeated. For example, if `period = Dates.Month(1)`
 and `repeat_date = Dates.Date(1993, 11)`, November 1993 will be repeated.
 
+!!! note
```
nefrathenrici:
```suggestion
!!! warn
```
This could be a warning instead but either is fine

nefrathenrici (approved):
Thank you, I think the concrete example will be helpful.

docs/src/inputs.md
```diff
@@ -137,6 +137,16 @@ the duration of the period that has to be repeated. The `repeat_date` defines
 what date range needs to be repeated. For example, if `period = Dates.Month(1)`
 and `repeat_date = Dates.Date(1993, 11)`, November 1993 will be repeated.
 
+!!! warn
+
+    While `PeriodicCalendar` is described as an extrapolation boundary condition,
+    it may also change the data used during the time period where data is available.
+    For example, take the case where data from 2000 to 2020 is available, and the
+    period provided to the `PeriodicCalendar` is the year 2000. The data from 2000
```
kmdeck: maybe it is the period= Dates.Year(1), and repeat_date = DateTime(2000)?

kmdeck (approved):
great changes 😅 thank you for clarifying!

PR #225, opened 2026-07-02
Cache on file reads and make copy of data

### LINKED ISSUE #221: Cache is not used for FileReaders.NCFileReader
The `FileReaders.read` function first check the cache and return that if the array exists. If there is a cache miss, then it load the data from the file and preprocess the data. It returns the result, but never cache the read.

https://github.com/CliMA/ClimaUtilities.jl/blob/84db8e471e07b06e3017e92db57ced3d503bda53/ext/NCFileReaderExt.jl#L249-L262

ph-kev:
closes #221 - In `NCFileReaderExt`, any file reads for time varying data was not cached. Additionally, ownership of the data was not passed properly to the caller which lead to the caller having access to data stored in the cache. Finally, when initializing a `NCFileReader`, a `NCDataset` file handle was eagerly initialized instead of lazily initialized.

### DIFF
--- a/ext/NCFileReaderExt.jl
+++ b/ext/NCFileReaderExt.jl
@@ -129,14 +129,13 @@ function FileReaders.NCFileReader(
 
     # Get dataset from global dictionary. If not available, open the new dataset and add
     # entry to global dictionary
-    dataset, open_varnames = get!(
-        OPEN_NCFILES,
-        file_paths,  # We map the collection of files to the dataset
+    # We map the collection of files to the dataset
+    dataset, open_varnames = get!(OPEN_NCFILES, file_paths) do
         (
             NCDatasets.NCDataset(file_path_to_ncdataset; aggtime_kwarg...),
             Set([varname]),
-        ),
-    )
+        )
+    end
     # push! will do nothing when file is opened for the first time
     push!(open_varnames, varname)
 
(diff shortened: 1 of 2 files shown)


ext/NCFileReaderExt.jl
```diff
@@ -129,14 +129,13 @@ function FileReaders.NCFileReader(
 
     # Get dataset from global dictionary. If not available, open the new dataset and add
     # entry to global dictionary
-    dataset, open_varnames = get!(
```
imreddyTeja, line 132: How did this work. in the first place? If not using the `do` syntax, shouldn't the first arg be a function?

ext/NCFileReaderExt.jl
```diff
@@ -280,11 +278,13 @@ function FileReaders.read(file_reader::NCFileReader)
         error("File contains temporal data, date required")
 
     # When there's no dates, we use DateTime(0) as key
-    return get!(file_reader._cached_reads, Dates.DateTime(0)) do
-        return file_reader.preprocess_func.(
-            Array(file_reader.dataset[file_reader.varname]),
-        )
-    end
+    return copy(
```
imreddyTeja, line 282: why is this a copy now

imreddyTeja (approved):
I'm a bit confused why some things now returns copies

ext/NCFileReaderExt.jl
ph-kev: Before, it was using this function signature: `get!(collection, key, default)`.

ext/NCFileReaderExt.jl
ph-kev: It is a copy, because we can't return a reference to the data in the cache. If we do return a reference, then the data can be mutated by the caller which corrupt the cache.

PR #226, opened 2026-07-03
Fix bugs and typos in tests

ph-kev:
I asked Claude to find bugs and typos in the tests and fix them. Then, I looked at each of one of the changes and confirm them.

### DIFF
--- a/test/data_handling.jl
+++ b/test/data_handling.jl
@@ -209,7 +209,6 @@ end
 
     var1 = "var1"
     var2 = "var2"
-    var3 = "var3"
 
     # Construct space
     FT = Float32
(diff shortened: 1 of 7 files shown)


test/data_handling.jl
```diff
@@ -438,8 +437,8 @@ end
                 ) == available_dates[11]
                 @test DataHandling.next_date(
                     data_handler,
-                    available_dates[10],
```
imreddyTeja, line 441: Why is this changed?

ph-kev: It is exactly the same as the test before it.

imreddyTeja: Yes... I was just wondering if there was a reason behind changing the indices used

imreddyTeja (approved):

test/data_handling.jl
ph-kev: It is either removing the test or updating the indices. I thought the test is interesting because `next_date` need to loop back and return the first index.

PR #228, opened 2026-07-06
Remove unnecessary allocations in read and read!

ph-kev:
On `main`, the current implementation of `read!` is calling `read` and using the array to fill out `dest`. For example, it looks like this:

```julia
dest .= FileReaders.read(file_reader, date)
```

This is not a good pattern because the right hand side still allocates a temporary array. This PR changes this by implementing `read` by allocating the destination array and calling `read!` with it.

### DIFF
--- a/ext/NCFileReaderExt.jl
+++ b/ext/NCFileReaderExt.jl
@@ -65,6 +65,9 @@ struct NCFileReader{
 
     """Index of the time dimension in the array (typically first). -1 for static datasets"""
     time_index::Int
+
+    """Size of the output array. This is used by read to initialize an array."""
+    output_size::Tuple{Vararg{Int}}
 end
 
 """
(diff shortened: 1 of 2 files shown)


ext/NCFileReaderExt.jl
```diff
@@ -228,36 +248,9 @@ end
...
-            index = index[]
-            var = file_reader.dataset[file_reader.varname]
-            slicer = [
-                i == file_reader.time_index ? index : Colon() for
-                i in 1:length(NCDatasets.dimnames(var))
-            ]
-            file_reader.preprocess_func.(var[slicer...])
-        end,
-    )
+    dest = valtype(file_reader._cached_reads)(undef, file_reader.output_size...)
```
imreddyTeja, line 251: why `valtype` instead of `eltype`

imreddyTeja (approved):

ext/NCFileReaderExt.jl
ph-kev: `file_reader._cached_reads` is an `AbstractDictionary`, so `valtype` return the type of the value which is what we want here. `eltype` would return something of the form `Pair{K, V}`.

PR #232, opened 2026-07-11
feat: add exact integer division for ITime

haakon-e:
Add `Base.:/(t::ITime, num::Integer)`

When the counter is divisible, the result keeps the current period and counter type. Otherwise the result takes the largest unit period (week down to nanosecond) at which the division is exact: `ITime(1, period = Second(1)) / 4` is 250 ms, and `ITime(1, period = Hour(1)) / 2` is 30 minutes. The division throws when no such period exists (`1 s / 3`) or the resulting counter does not fit the counter type; division by zero throws `DivideError`. `div(t, num)` is unchanged: it truncates at the current period, as `div` does for other numeric types.

**Motivation:** sub-stepping wants `dt / n_sub` exact so every sub-step is identical[^1]. 
`div` truncates: `div(ITime(10, period = Second(1)), 4)` is 2 s, and a sub-cycle built on it covers 8 of the 10 seconds, while `ITime(10, period = Second(1)) / 4` is exactly 2500 ms. 

An invalid count (the same step divided by 3) throws at configuration time instead of silently undershooting. First use case: the acoustic-substepping driver behind CliMA/ClimaTimeSteppers.jl#442 and CliMA/ClimaAtmos.jl#4660.

[^1]: We could instead absorb the division remainder by extending one sub-step

**Verification:** the `test/itime.jl` testset covers division at the current period, refinement from every unit period, multiple-unit periods (`Millisecond(100)`, `Hour(2)`), negative counters and divisors, a division beyond `typemax(Int64)` nanoseconds that succeeds exactly (500 years of 365 days, halved), the non-representable, overflow, and divide-by-zero errors, and epoch and counter-type preservation (`Int32` stays `Int32`; `UInt64` and `BigInt` are exercised).

The behavior is documented on the ITime documentation page, and the docstring is listed in the TimeManager API block. `seconds` also gains a `Real` method (the value is taken to already be in seconds), so call sites handling either time representation can use the unit-explicit name instead of `float`.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ClimaUtilities"
 uuid = "b3f4f4ca-9299-4f7f-bd9b-81e1242a7513"
-version = "0.1.30"
+version = "0.1.31"
 authors = ["Gabriele Bozzola <gbozzola@caltech.edu>", "Julia Sloan <jsloan@caltech.edu>", "Kevin Phan <kphan2@caltech.edu>"]
 
 [deps]
--- a/src/ITime.jl
+++ b/src/ITime.jl
@@ -474,5 +474,79 @@ Base.:*(num::Integer, t::T) where {T <: ITime} =
 Base.:*(t::T, num::Integer) where {T <: ITime} =
     ITime(num * t.counter, t.period, t.epoch)
 
+"""
+    finer_unit_period(period::Dates.FixedPeriod)
+
+Return the next unit period after the unit of `period` in the sequence week,
+day, hour, minute, second, millisecond, microsecond, nanosecond, as a pair
+`(finer_period, ratio)` with `oneunit(period) == ratio * finer_period`.
+Return `nothing` when the unit of `period` is `Dates.Nanosecond`.
+"""
+finer_unit_period(::Dates.Week) = (Dates.Day(1), 7)
+finer_unit_period(::Dates.Day) = (Dates.Hour(1), 24)
+finer_unit_period(::Dates.Hour) = (Dates.Minute(1), 60)
+finer_unit_period(::Dates.Minute) = (Dates.Second(1), 60)
+finer_unit_period(::Dates.Second) = (Dates.Millisecond(1), 1000)
+finer_unit_period(::Dates.Millisecond) = (Dates.Microsecond(1), 1000)
+finer_unit_period(::Dates.Microsecond) = (Dates.Nanosecond(1), 1000)
+finer_unit_period(::Dates.Nanosecond) = nothing
+
+"""
+    refine_period(period::Dates.FixedPeriod)
+
+Return the next step in the period-refinement sequence as a pair
+`(finer_period, ratio)` with `period == ratio * finer_period`: a multiple-unit
+period refines to its unit period, and a unit period to the next finer unit
+period. Return `nothing` for `Dates.Nanosecond(1)`.
+"""
+function refine_period(period::Dates.FixedPeriod)
+    n = Dates.value(period)
+    n == 1 || return (typeof(period)(1), n)
+    return finer_unit_period(period)
+end
+
+"""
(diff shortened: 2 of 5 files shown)


src/ITime.jl
```diff
@@ -474,5 +479,45 @@ Base.:*(num::Integer, t::T) where {T <: ITime} =
...
+
+Unlike `div(t, num)`, which truncates the counter, this division is exact.
+"""
+function Base.:/(t::ITime, num::Integer)
+    iszero(num) && error("Cannot divide ITime by zero")
+    INT = typeof(t.counter)
+    iszero(rem(t.counter, num)) &&
+        return ITime(INT(div(t.counter, num)), t.period, t.epoch)
+    current_ns = Dates.tons(t.period)
+    total_ns = Int128(t.counter) * Int128(current_ns)
```
ph-kev: Could you write a test case where this is actually needed? I am guessing Int128 are used to protect against overflow, but I am not sure? If you are worried about overflow, you should check for overflow instead.

src/ITime.jl
```diff
@@ -474,5 +479,45 @@ Base.:*(num::Integer, t::T) where {T <: ITime} =
...
+    for finer_period in (
+        Dates.Second(1),
+        Dates.Millisecond(1),
+        Dates.Microsecond(1),
+        Dates.Nanosecond(1),
```
ph-kev: Is there a reason why we don't include `Dates.Week`, `Dates.Day`, `Dates.Hour`, and `Dates.Minute`?

test/itime.jl
```diff
@@ -145,6 +145,53 @@ using Test, Dates
...
+        @test_throws ErrorException ITime(1, period = Dates.Second(1)) / 0
+
+        # Epoch preservation
+        epoch = Dates.DateTime(2020, 1, 1)
+        t4 = ITime(1, period = Dates.Second(1), epoch = epoch)
+        @test (t4 / 2).epoch == epoch
+        @test (t4 / 4).epoch == epoch
+
+        # Counter type preservation
+        t5 = ITime(Int32(4), period = Dates.Second(1))
```
ph-kev, line 261: This behavior should be documented as part of the user facing documentation.

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "ClimaUtilities"
 uuid = "b3f4f4ca-9299-4f7f-bd9b-81e1242a7513"
-version = "0.1.29"
+version = "0.1.30"
```
ph-kev: If you are going to bump the version, then the NEWS shouldn't be under main, but under v0.1.30.

ph-kev (changes requested):

src/ITime.jl
haakon-e: Right - it was overflow protection: I computed the intermediate value in nanoseconds, which overflows Int64 for any time longer than about 292 years, at any period. No one is dividing times that long[^1], so I removed the `Int128` conversion (though I still compute in at least `Int64` internally). Instead, I iteratively search for the coarsest period at which the division is exact, cancelling common factors of the divisor along the way and checking each step with `checked_mul` from Julia Base. Overflow can now only happen if the result itself is not representable in the counter type.

[^1]: some pre-industrial control runs do exceed this; the new version handles times of that length (there is a test dividing 500 years), since it no longer goes through a nanosecond representation.

src/ITime.jl
haakon-e: No good reason - I wrote the code for the case I had in mind (a timestepper's `dt`, which is (usually) not coarser than seconds). I've added all the units, week through nanosecond, with tests.

test/itime.jl
haakon-e: Done. Let me know if you have any feedback.

Project.toml
haakon-e: thanks! Moved to the right location.

PR #239, opened 2026-07-21
Fix premature close in NetCDF dataset registry

### LINKED ISSUE #238: Premature close for multiple readers
If two readers share the same dataset and one of the reader is closed, then the other reader will also not be able to read the dataset. This is because the registry

```julia
const OPEN_NCFILES =
    Dict{Union{String, Vector{String}}, Tuple{NetCDFDataset, Set{String}}}()
```

doesn't keep track of how many readers is using the same dataset and variable.

ph-kev:
closes #238 - For the NCFileReader, multiple readers can share the same dataset. If one reader was closed, the shared dataset was closed, so a still-open reader could no longer read from it.

### DIFF
--- a/test/file_readers.jl
+++ b/test/file_readers.jl
@@ -7,6 +7,8 @@ import ClimaUtilities.FileReaders
 using NCDatasets
 
 @testset "NCFileReader with time" begin
+    # Start from a clean OPEN_NCFILES state
+    FileReaders.close_all_ncfiles()
     PATH = joinpath(artifact"era5_example", "era5_t2m_sp_u10n_20210101.nc")
     NCDataset(PATH) do nc
         ncreader_sp = FileReaders.NCFileReader(PATH, "sp")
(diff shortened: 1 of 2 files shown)


test/file_readers.jl
```diff
@@ -66,8 +66,40 @@ using NCDatasets
...
-        length(FileReaders.available_dates(ncreader_agg)) == 104
+        @test FileReaders.available_dates(ncreader_agg) == nc["time"][:]
+        @test length(FileReaders.available_dates(ncreader_agg)) == 104
+        close(ncreader_agg)
+    end
+end
+
+@testset "Shared readers of the same variable" begin
+    PATH = joinpath(artifact"era5_example", "era5_t2m_sp_u10n_20210101.nc")
+    open_ncfiles =
```
imreddyTeja, line 80: Will this test potentially fail if the const `OPEN_NCFILES` is polluted when the the test is ran? I remember this was an issue with the ClimaLand diagnostics tests, and it made it very annoying to run the tests interactively.

imreddyTeja (commented):
Changes look good to me, but I left one comment about a concern about the tests

test/file_readers.jl
ph-kev: I don't think there is an issue if you run the tests in order, but if new tests are added, I can see the possibility of it failing. I decided to add `FileReaders.close_all_ncfiles()` to the start of the testsets.

imreddyTeja (approved):

PR #241, opened 2026-07-22
Add support for multiple columns interpolation

ph-kev:
This is building on top of https://github.com/CliMA/ClimaCore.jl/pull/2525 which add support for spaces with multiple columns. To complete https://github.com/CliMA/ClimaLand.jl/issues/1723, this PR adds support for loading data from datasets which each dataset represents a single column. In addition to this, regridding along the vertical direction for this kind of data is also added in this PR. Note that for ClimaLand, the forcing data for the single column runs is only on the surface which means that no interpolation is done and only a memcopy is done. An example of this working in ClimaLand can be found [here](https://github.com/CliMA/ClimaLand.jl/tree/kp/multi-cols) (note that most of the code there is generated by LLM and it is more of a proof of concept than something that is ready to be merge).

Note that most of the code generated in this PR is done by LLM and verified by me.

# Design Notes

Here's a list of brief design considerations and notes for each struct.

## `ColumnRegridder`

This is needed since `InterpolationsRegridder` can't handle regridding non-gridded source data. Note that you can still use `InterpolationRegridder` with the new spaces since all the ClimaCore spaces are non-gridded. 

## `DataSource` 
This struct exists to simplify the logic of the input arguments to MultiColumnNCFileReader. Otherwise, this would be a vector of vectors of vector of vectors of datasets representing columns. The each level of the vectors correspond to time concatenation of datasets, different variables, and different columns.

This is also needed to support time preprocessing which is not currently available. For ClimaLand, each column has its own UTC offset, so a single time preprocessing function can't be used here. It must be done for each column.

One thing that isn't done is storing the dimension values which I am not sure if it should be added or not.

In the future, this can be used to support other types of datasets (not just NetCDF files) but would require a bit more functionality such as opening and closing datasets. Also, `DataSource` can also be used in the `NCFileReader` but it is not done in this PR to avoid more changes.

## `MultiColumnNCFileReader`
 
The different requirements than the `NCFileReader` with the kind of checks that we need to do and the reading of files is slightly different. I don't think it is possible to adapt `NCFileReader` to make it work with the kind of data that we are handling.

## `MultiColumnDataHandler`

The fields of this struct are identical to the `DataHandler`. One way to avoid code duplication is to use a trait system (`e.g. `Column` vs `Gridded`). Multiple dispatch can't be used here since the current list of arguments doesn't give anything to dispatch off of.

This struct should be able to support spaces that have a single column, but it does not right now to simplify the code that need to be written.

## Other considerations

I haven't benchmark this code against the existing code, but I suspected that reading from N datasets can be slow. It might be worth looking into a ring buffer instead of a LRU cache and make the cache size big enough to store all the data of the datasets before the simulation began.

# Questions

1. Columns are chosen at the level of the `MultiColumnDataHandler`. I am not sure if this is the best place to do this kind of thing or if the feature should be added. It introduces a lot of complexity to the code, since we need to open the datasets and look for this kind of information (although this can be done when making the `DataSource` objects). In addition, it imposes a requirement that datasets have that information which I am not sure that all datasets have. However, it does make it convenient for the users and remove possible user errors if the users didn't list the columns in the right order. 
2. Are there datasets that don't have a longitude and latitude directions supported? Is this something we need to support if this will be used in ClimaAtmos single column runs? Maybe there need to be a feature to disable this check?

# TODO

- [ ] User facing documentation for all structs
- [x] Check GPU compatibility

### DIFF
--- a/ext/ClimaUtilitiesClimaCoreClimaInterpolationsExt.jl
+++ b/ext/ClimaUtilitiesClimaCoreClimaInterpolationsExt.jl
@@ -1,0 +1,5 @@
+module ClimaUtilitiesClimaCoreClimaInterpolationsExt
+
+include("ColumnRegridderExt.jl")
+
+end
--- a/ext/ColumnRegridderExt.jl
+++ b/ext/ColumnRegridderExt.jl
@@ -1,0 +1,236 @@
+module ColumnRegridderExt
+
+import ClimaInterpolations
+
+import ClimaCore
+import ClimaComms
+
+import ClimaUtilities.Regridders
+
+"""
+    ColumnRegridder
+
+A regridder that interpolates data on each column's vertical levels onto the
+vertical levels of `target_space`, using `ClimaInterpolations.jl`. If there are
+no vertical levels, then no vertical interpolation is done.
+
+Unlike `InterpolationsRegridder`, this does no horizontal interpolation. The
+columns are assumed to already be aligned with the target space's columns.
+"""
+struct ColumnRegridder{
+    SPACE <: ClimaCore.Spaces.AbstractSpace,
+    OR,
+    EX,
+    SRC <: AbstractArray,
+    TARGET <: AbstractArray,
+    SCR <: Union{AbstractArray, Nothing},
+} <: Regridders.AbstractRegridder
+    """ClimaCore space to interpolate onto."""
+    target_space::SPACE
+
+    """Interpolation order (e.g.
+    `ClimaInterpolations.Interpolation1D.Linear()`)."""
+    order::OR
+
+    """Extrapolation condition (e.g.
+    `ClimaInterpolations.Interpolation1D.Flat()`)."""
+    extrapolate::EX
+
+    """Source vertical levels whose size is `n_source_levels` x `n_columns`.
+    Empty when there is no vertical dimension."""
+    src_vertical_levels::SRC
+
+    """Vertical levels of the target space. Empty when there is no vertical
+    dimension."""
+    target_vertical_levels::TARGET
+
+    """Whether the data passed to `regrid` must be reversed along the vertical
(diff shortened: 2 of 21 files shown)


ext/ClimaUtilitiesClimaCoreClimaInterpolationsExt.jl
nefrathenrici, line 1: Is there a reason we shouldn't just make ClimaInterpolations a full dependency? It's a pretty lightweight package.

ext/ColumnRegridderExt.jl
```diff
@@ -0,0 +1,236 @@
...
+"""
+function Regridders.regrid(regridder::ColumnRegridder, data)
+    AT = ClimaComms.array_type(regridder.target_space)
+
+    field = zeros(regridder.target_space)
+    target_array = ClimaCore.Fields.field2array(field)
+
+    # This can happen with surface data (e.g. LAI data in ClimaLand)
+    if isempty(regridder.src_vertical_levels)
+        copyto!(target_array, vec(data))
```
nefrathenrici, line 212: This should include a size check as is done elsewhere in this function, since if `length(src) < length(dest)` it will leave zeros/existing data in the dest.

ext/DataSourceExt.jl
```diff
@@ -0,0 +1,200 @@
+module DataSourceExt
+
+import ClimaUtilities.FileReaders
+
+import Dates
+import NCDatasets
+
+include("nc_common.jl")
```
nefrathenrici, line 8: This is now called three times in the package, maybe this should be made a module?

nefrathenrici (commented):
Looks good, this is a big PR so I only gave partial review. I'll take a closer look tomorrow.

nefrathenrici (approved):
LGTM, I didn't find any other clear bugs in my second review

PR #242, opened 2026-07-24
Use binary search instead of linear scans

ph-kev:
There are parts of the code that does a linear scan on `available_dates`. This is verified to be sorted at construction. As such, I updated the code to do a binary search instead of a linear scan whenever possible.

I asked a LLM to implement these changes.


imreddyTeja (approved):
I didn't know base Julia has these functions

nefrathenrici (approved):

PR #243, opened 2026-08-03
Fix double close bug with NetCDF dataset registry

ph-kev:
This PR fixes a double close bug with the NetCDF dataset registry. The bug is that you can call `close` multiple times on a single readers and it would close the NetCDF file when the count hits zero. This can leads to a reader not being able to read a file if there are two readers reading the same file and `close` was called multiple time on one of the reader.

To fix this, we make `close` idempotent by keeping track of the open readers.

### DIFF
--- a/ext/NCFileReaderExt.jl
+++ b/ext/NCFileReaderExt.jl
@@ -70,6 +70,9 @@ struct NCFileReader{
 
     """Size of the output array. This is used by read to initialize an array."""
     output_size::Tuple{Vararg{Int}}
+
+    """Whether this reader has not been closed yet. Used to make `close` idempotent."""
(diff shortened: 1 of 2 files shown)


imreddyTeja (approved):

imreddyTeja:
Why are the downstream tests failing?

ph-kev:
> Why are the downstream tests failing?

I think this is because of a memory leak. I thought `IdSet` use weak references but they don't. I changed the solution to add a field that keep a reference to a boolean that track if the reader is opened or not. Although, this introduces some complication with closing all the NetCDF files.

imreddyTeja (approved):

PR #244, opened 2026-08-03
Update ClimaCore extension

dennisYatunin:
## Purpose 
This PR updates `TempestRegridderExt.jl` following CliMA/ClimaCore.jl#2522. Every DataLayout used in ClimaAtmos is now a VIJFH, so the parent array manipulation needed to be changed.

----

### DIFF
--- a/ext/TempestRegridderExt.jl
+++ b/ext/TempestRegridderExt.jl
@@ -119,40 +119,86 @@ function Regridders.regrid(regridder::TempestRegridder, date::Dates.DateTime)
     )
 end
 
-"""
-    reshape_cgll_sparse_to_field!(field::Fields.Field, in_array::Array, R)
+@static if pkgversion(ClimaCore) >= v"0.15.0"
+    """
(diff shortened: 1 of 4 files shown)


ph-kev (approved):
It would be nice if we can merge https://github.com/CliMA/ClimaUtilities.jl/pull/243 before merging this PR, but I am fine with it either way.

ext/TempestRegridderExt.jl
```diff
@@ -138,28 +138,20 @@ function reshape_cgll_sparse_to_field!(
     in_array::Array,
     R,
 )
-    field_array = parent(field)
-
-    fill!(field_array, zero(eltype(field_array)))
-    Nf = size(field_array, 3)
+    target = ClimaCore.Fields.field_values(field)
```
ph-kev: If we want to support both v0.14 and v0.15, we can use `@pkgversion`. This is up to you though.

imreddyTeja:
> It would be nice if we can merge #243 before merging this PR, but I am fine with it either way.

Good idea. I'll review it now

ph-kev (approved):

PR #245, opened 2026-08-07
Refactor TimeVaryingInputs0DExt

ph-kev:
While trying to add support for interpolation for multiple points/columns, I needed to extend on `TimeVaryingInputs0DExt` which was difficult to do given the current code. This PR refactors all of it using https://github.com/CliMA/ClimaAnalysis.jl/pull/418 as the basis for the refactor.

I used a LLM to write the code and I verified it.

### DIFF
--- a/ext/TimeVaryingInputs0DExt.jl
+++ b/ext/TimeVaryingInputs0DExt.jl
@@ -1,36 +1,30 @@
 module TimeVaryingInputs0DExt
 
-import ClimaCore
-import ClimaCore: ClimaComms
-import ClimaCore.Fields: Adapt
+import Dates: DateTime
 
-import ClimaUtilities.Utils:
-    searchsortednearest, linear_interpolation, wrap_time, isequispaced
+import ClimaUtilities.Utils: searchsortednearest, wrap_time, isequispaced
+import ClimaUtilities.TimeVaryingInputs
 import ClimaUtilities.TimeVaryingInputs:
-    AbstractInterpolationMethod, AbstractTimeVaryingInput
-import ClimaUtilities.TimeVaryingInputs:
+    AbstractInterpolationMethod,
+    AbstractTimeVaryingInput,
     NearestNeighbor,
     LinearInterpolation,
     LinearPeriodFillingInterpolation,
     Throw,
     Flat,
-    PeriodicCalendar
-import ClimaUtilities.TimeVaryingInputs: extrapolation_bc
-
-import ClimaUtilities.TimeVaryingInputs
-
+    PeriodicCalendar,
+    extrapolation_bc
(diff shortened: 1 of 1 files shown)


ext/TimeVaryingInputs0DExt.jl
```diff
@@ -54,33 +48,6 @@ struct InterpolatingTimeVaryingInput0D{
...
-
-    TimeVaryingInputs.evaluate!(scalar_dest, itp, time, itp.method)
-    fill!(destination, scalar_dest[])
-
-    return nothing
-end
-
 function TimeVaryingInputs.TimeVaryingInput(
     times::AbstractArray,
     vals::AbstractArray;
```
ph-kev, line 53: Do we know why `times` and `vals` can be `AbstractArray`? The only use case I can think of is if `times` and `vals` are `1 x N` matrices. I think it would be simpler if they are `AbstractVector`s.

ext/TimeVaryingInputs0DExt.jl
```diff
@@ -96,9 +63,37 @@ function TimeVaryingInputs.TimeVaryingInput(
...
+end
+
+"""
+    _validated_times(times, method)
+
+Check that `times` is sorted and compatible with `method`, promoting `ITime`s
+to a common epoch and period when needed and return the validated times.
+"""
+function _validated_times(times, method)
+    issorted(times, lt = <=) ||
```
imreddyTeja, line 95: Is this behavior change needed and intentional? If so, the other uses of `issorted` should be updated to match

ext/TimeVaryingInputs0DExt.jl
```diff
@@ -112,194 +107,202 @@ function TimeVaryingInputs.TimeVaryingInput(
...
+"""
+function Base.in(time, itp::InterpolatingTimeVaryingInput0D)
+    return itp.range[1] <= time <= itp.range[2]
 end
 
-function TimeVaryingInputs.evaluate!(
-    dest,
-    itp::InterpolatingTimeVaryingInput0D,
-    time,
-    ::NearestNeighbor,
```
imreddyTeja, line 171: nit: If these are considered public functions, then this is a breaking change.

Have you checked nothing downstream explicitally calls `evaluate!` with an interpolation method

imreddyTeja (approved):

ext/TimeVaryingInputs0DExt.jl
ph-kev: The behavior change is intentional since it doesn't match the docstring. Also, I would say that repeated times shouldn't happen in the first place.

I don't know what other use of `issorted` that you are talking about unless it is in a different file?

